PR #13, opened 2018-07-21
Enable plugins

mateosalta:
This pull enables support for plugins, with this a user can more easily add plugins such as flash or widevine to the browser.

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -26,6 +26,9 @@ import "actions" as Actions
 WebEngineView {
     id: webview
     property var currentWebview: webview
+    
+    //enable using plugins, such as widevine or flash, to be installed separate
+    settings.pluginsEnabled: true
 
     property QtObject __ua: UserAgent02 {
     }


mateosalta:
Also note that plugins also need a launch flag, the only one that is recognized in its location without a flag is flash, if this is a concern we could have this as a settings option. But I think this is an imortant step towards flash or widevine(netflix)

Flohack74:
IMHO Flash should be used by noone anymore these days ;)

mateosalta:
True. there are still some hold-outs, and I like that we can make this a possibility. And a step towards setting up netflix.

(side note, on of my favorite sites uses flash http://homestarrunner.com/ and it would be really nice to add to the "iphone can't do that" list)

NeoTheThird:
> IMHO Flash should be used by noone anymore these days ;)

Yeah, but we should still allow it. And making it easy to use widevine will be very interesting for many users!

@mariogrip review please.

mariogrip:
Thanks :+1:

Issue #5, opened 2018-06-04
white area overlaying browser


doniks:
repro:
open forums.ubports.com
tab into any text field, like login, or search

observe:
there is a white area blocking out part of the browser.
the size of the area looks suspiciously identical to the size of the keyboard

device:
nexus 7 deb
xenial
![keyboard](https://user-images.githubusercontent.com/26801304/40930920-659fcf00-6829-11e8-915f-d884492e52f7.png)

doniks:
happens on all textfields on all websites 
happens also when entering URL
does NOT happen in desktop mode!

Trim:
Hello,

I've just installed 16.04 W26 and the browser-ng: I notice the same issue on an aquaris m10 (cooler) device.

For me, it's like if the white part should be place for the keyboard.

It seems for me that the OSK is a little bit more compact on 16.04, maybe, the issue is just about the expected heigh of the OSK ?

hummlbach:
The same on mako. Unfortunately here the osk takes half of the screen s.t. the other half is blank.

advocatux:
I can reproduce this issue too on BQ E4.5 (krillin) running UT 16.04 (2018-07-01), with the only difference that it's even worse on this device

![browser-ng](https://user-images.githubusercontent.com/30233652/42163371-82de0638-7e02-11e8-923f-49665babdcdb.jpg)

ruedigerkupper:
This really is a show stopper. It renders browser next unusable.

mateosalta:
seems like the keyboard avoidance went wrong somewhere, I fixed by removing this on the youtube webapp I'm working on - but then content does flow under the keyboard (nice for transparent keyboard) this is ok for youtube, but I know some sites the keyboard will cover the textfield. 

Would this situation be better than the white box?

cyberpunkedu:
I'm getting something similar in Morph in webapps using the ogra container. The page loads, then it goes all white, but it seems unrelated to the keyboard.

ruedigerkupper:
Confirming the problem that userj reports (webapps using the ogra container turn completely white after loading). But I'm not sure it is really the same thing as the white keyboard overlay.

mateosalta:
yeah, they are crashing because they are on the old browser using oxide. not related.

cyberpunkedu:
> yeah, they are crashing because they are on the old browser using oxide. not related.

I modified the container to use Qtwebengine
`import QtWebEngine 1.7 as Morph`

[...]

PR #15, opened 2018-07-22
Temporary Keyboard overlay fix

nfsprodriver:
... completely removing white space

### DIFF
--- a/src/app/KeyboardRectangle.qml
+++ b/src/app/KeyboardRectangle.qml
@@ -28,7 +28,7 @@ Item {
     anchors.left: parent.left
     anchors.right: parent.right
     anchors.bottom: parent.bottom
-    height: Qt.inputMethod.visible ? Qt.inputMethod.keyboardRectangle.height : 0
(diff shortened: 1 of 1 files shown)


nfsprodriver:
#5

NeoTheThird:
I'm wondering what the rationale for the rectangle was in the first place. It might be nice not to have it, if you want keyboard opacity, for example.

NeoTheThird:
Ah!

> This file was originally part of the telephony application.
It is a workaround required to handle interaction with the OSK,
until the shell/WM takes care of that on behalf of the applications.

https://github.com/ubports/browser-ng/pull/15/files#diff-1c9bca71438f627d69c9571b4f62bd76R19

NeoTheThird:
Isn't what we're seeing here that the shell is actually taking care of it, and the workaround is just messing everything up? If so, we can just remove this hack entirely, not just make it invisible :)

Pinging @mariogrip for Mir and Unity8 stuff.

NeoTheThird:
Last edits (except from updating to newer libs) were 5 years ago.

mariogrip:
Thanks :+1:

PR #23, opened 2018-08-24
change user agent (recaptcha page stated "not supported browser")

balcy:
we should check the change has no unwanted side effects

tries to solve https://github.com/ubports/browser-ng/issues/22


balcy:
we can close this one, and I can create a new PR for "morph-browser" (new Morph library)
the branch for the change after the rename was done is
https://github.com/balcy/morph-browser/tree/xenial_-_morphuseragent

PR #25, opened 2018-09-11
Remove scale factor 2 due to latest changes

nfsprodriver:


UniversalSuperBox:
A couple of lines after you set that environment variable, this is done:

`QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);`

Only one or the other is needed, not both. However, the build is failing due to that `setAttribute`, so maybe you want to see why that's happening?

UniversalSuperBox:
Never mind, it's due to the branch naming. Either way, this doesn't need to be set twice. Just a removal of the environment variable setting should be enough.

Issue #29, opened 2018-09-17
enable downloads


balcy:
I've started to work on this in
https://github.com/balcy/browser-ng/tree/xenial_-_downloads
but have not yet filed a PR.

Since QtWebengine has its own downloads API (https://doc.qt.io/qt-5/qml-qtwebengine-webengineprofile.html#downloadRequested-signal) the Ubuntu Download Manager is no longer used.

It is already working for the downloads of the current session, but the downloads page does not show the already completed downloads of previous sessions. (maybe sth. not correct with the SQLite database of the downloadsModel (downloads-model.cpp) regarding the new implementation.)

One thing I've seen is that the QML downloads object does not include the URL property, while the C++ class can access the URL as readonly function:
I've asked that here: https://forum.qt.io/topic/94571/webenginedownloaditem-url-property
Or am I missing something ?

UniversalSuperBox:
Looks like downloads are working now!
![screenshot20181008_163140227](https://user-images.githubusercontent.com/21966173/46634789-aeac1680-cb17-11e8-8602-2e67e115ad7c.png)

PR #32, opened 2018-09-22
enable downloads (links, images, media), fix click builds

balcy:
https://github.com/ubports/morph-browser/issues/29


mariogrip:
Awesome! LGTM :+1:

balcy:
thanks ! One thing to note is that Ubuntu Download manager is not used, but the builtin mechanism of QtWebEngine. So I guess downloads will only be running when the app has the focus, and not in the background.
but without getting the URL (https://forum.qt.io/topic/94571/webenginedownloaditem-url-property) there seems to be no straight-forward way for it. But let's see...

Issue #22, opened 2018-08-24
Recaptcha page says "browser not supported"


balcy:
recaptcha did not accept browser-ng (states that it is not a supported browser)
there is an example page to test recaptcha:
https://patrickhlauke.github.io/recaptcha/

It seems to be interpret the user agent and block the browser because of that.
We can try another user agent string, but should avoid side effects like pages directing users to the Play Store (because they think the browser is android) or wrong rendering of pages...

balcy:
https://github.com/balcy/morph-browser/tree/xenial_-_morphuseragent (for the new morph-browser)

balcy:
it is really a pity that one cannot reach companies directly to report issues nowadays. There are several chromium related posts in the "support" forum, one Ubuntu Touch related one of 2016.
https://groups.google.com/forum/m/#!searchin/recaptcha/chromium/recaptcha/e_QdyJR36_k
I think it would be better not using recaptcha at all, but you cannot decide as user :[
Maybe we can make it so that we present the Chrome user agent to recaptcha servers only, and keep Chromium generally.

dobey:
Odd. I've been using Chromium exclusively for years on my Ubuntu systems, and never once had an issue with recaptcha due to the user agent. So very odd that changing it to Chrome here would have any effect. Perhaps it's more related to the fact that we stick "Android" into the user agent as a means of tricking web sites to serve up mobile versions, and perhaps Google does require Chrome in that case?

Stefano0101:
Yes, this issue is confirmed on Meizu Pro 5. 16/04/OTA-5/stable.

Billdemoncho:
I confirm the same problem on my meizu pro 5 in 16.04 dev and updated.

![screenshot20181016_234417243](https://user-images.githubusercontent.com/42817545/47050110-7fdd1280-d19f-11e8-9d3a-1b4850983693.png)

ghost:
device:Nexus 5
channel: RC 2018-W48
issue solved. Unable to reproduce using recaptcha test linked above.

tjrhodes:
Device: FP2
System Build: 2018-W48

Issue solved, the test link loaded the vanilla recaptcha widget and it was functional.

As mentioned above though it is odd, Chromium on Ubuntu desktop has no such problems, is it confirmed to be related to putting android in the request headers? Can anyone test with chromium on android?

PR #33, opened 2018-09-22
change user agent

balcy:
for https://github.com/ubports/morph-browser/issues/22


balcy:
note: while it does solve the recaptcha problem, some pages seem to redirect the users to Google Play store (e.g email provider GMX) with a popup window. I have not found a user agent yet that does not mention android but works with recaptcha.

balcy:
ok found one way: if I use the original user agent (current ubports xenial branch), and change Chromium to Chrome, it works as well. But that's a bit against my philosophy, isn't it : )

PR #36, opened 2018-10-03
Hide scrollbar

mateosalta:
by enabling userscript, hide the builtin scrollbar, that is only themeable via css.


mariogrip:
Could you add this in the new Morph library so all apps can take use of this? 
https://github.com/ubports/morph-browser/blob/xenial/src/Morph/Web/MorphWebContext.qml#L49

mariogrip:
Sorry if i'm a little annoying moving you around first to xenial_-_morph now this :)

mateosalta:
cool, moved to xenial_-morph. ah, I see - to the global scripts, then even embeded views can take advantage of this, and further themes of the scroll bar, right? Awesome, I should be able to move it there easily.

mateosalta:
I think this is right, but haven't been able to test, as I'm not having luck rebuilding.

PR #38, opened 2018-10-09
add selection of element contents (e.g. text of current div) and its parent(s)

balcy:
Currently if you long-press on a page, the current word gets selected, and the context menu offers "select all".
So you can either copy/share the word or the whole page. (for inputs/textareas it is better because you can select the text in that with "select all")
With this PR a new little feature "select element" is added, that allows to select the text / content of the element the context menu was opened in.

A typical use case for me is copying build / install instructions from a github / gitlab about page, and pasting them into the terminal app.
examples I've used for testing:
https://github.com/benni0815/SchafKopf
http://docs.ubports.com/en/latest/userguide/dailyuse/libertine.html

Another case would be selecting one paragraph in a text only.

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -537,7 +537,7 @@ WebView {
                     name: "cut"
                     text: i18n.dtr('ubuntu-ui-toolkit', "Cut")
                     iconName: "edit-cut"
-                    enabled: contextMenuRequest && contextMenuRequest.isContentEditable && (contextMenuRequest.editFlags & ContextMenuRequest.CanCut) && quickMenu.selectedTextLength > 0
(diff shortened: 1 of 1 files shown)


balcy:
some screenshots:
(1)
step_1
(2)
step_2
(3)
step_3

balcy:
why do you need several levels at all ? First I just implemented select element, but sometimes like the mentioned ubports docs page, a line consists of several pre elements.
Then at first I just made "select more" to go up exactly 1 parent, but like that not always more text is selected (invisible nesting). 
Therefore it always checks if the selection gets bigger, and remembers the history so that "select less" takes back the last action. (the user sees a visible change for each click)

PR #40, opened 2018-10-09
replace "Chromium" with "Chrome" (user agent)

balcy:
Chromium seems unsupported browser for ReCaptcha

for https://github.com/ubports/morph-browser/issues/22

 if I use the original user agent (current ubports xenial branch), and change Chromium to Chrome, it works as well. But that's a bit against open source philosophy, isn't it ?

But seems like a workaround for now. Can the Chrome entry have negative side effects ? E.g. Sites thinking features are available that exist for Chrome but not Chromium ?

### DIFF
--- a/src/Morph/Web/UserAgent02.qml
+++ b/src/Morph/Web/UserAgent02.qml
@@ -1,9 +1,9 @@
 /*
  * Copyright 2013-2016 Canonical Ltd.
  *
- * This file is part of webbrowser-app.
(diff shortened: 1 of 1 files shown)


balcy:
strange that it did not build, could the fact that 2 CI builds were running simultaneously cause that?

balcy:
ok just triggered rebuild without code change

Flohack74 (approved):

PR #48, opened 2018-10-12
Move browser data on first start of unity8

UniversalSuperBox:
This PR adds an Upstart job to move bookmarks, history, and downloads into morph-browser's directory on the start of unity8, if the `~/.local/share/morph-browser/`  directory doesn't exist.

### DIFF
--- a/debian/copy-browser-files.conf
+++ b/debian/copy-browser-files.conf
@@ -1,0 +1,19 @@
+description "Move browser-app files"
+author "Dalton Durst <dalton@ubports.com>"
+
+start on starting unity8
+
+script
+    set -x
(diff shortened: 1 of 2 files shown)


debian/copy-browser-files.conf
```diff
@@ -0,0 +1,19 @@
...
+
+start on starting unity8
+
+script
+    set -x
+    SOURCE_DIR="$HOME/.local/share/webbrowser-app/"
+    DEST_DIR="$HOME/.local/share/morph-browser/"
+
+    if [ ! -d $DEST_DIR ]; then
+        echo ""
```
mariogrip: huh?

mariogrip (approved):
LGTM :+1: but why is that echo there?

Issue #56, opened 2018-10-16
URL input box should not have autocorrect


UniversalSuperBox:
In the browser, the URL bar currently has autocorrect enabled:

![screenshot20181016_074606182](https://user-images.githubusercontent.com/21966173/47017464-f33b4180-d117-11e8-98c7-bd1a2ac49d7f.png)

This causes several problems, for example:

* Input isn't actually passed to the box until the keyboard is finished with the word, so autocomplete doesn't start until the user selects an autocorrect word or presses Space
* URLs aren't normally perfect for the language you're native to. The keyboard "helpfully" trying to correct them is annoying ("google.com can become Gogol. com")

## How to fix it

Set the [inputMethodHint](https://api-docs.ubports.com/sdk/apps/qml/Ubuntu.Components/TextField.html#sdk-ubuntu-components-textfield-inputmethodhints) to `Qt.ImhNoPredictiveText` in [AddressBar.qml](https://github.com/ubports/morph-browser/blob/xenial/src/app/webbrowser/AddressBar.qml). Note that you can set multiple input hints on one widget by adding a bar (`|`) between the values.

ghost:
device:Nexus 5
channel: RC 2018-W48
issue solved. Unable to reproduce

Ingo-FP-Angel:
device: FP2
channel: RC 2018-W48
Issue solved. Unable to reproduce.

PR #58, opened 2018-10-16
Removed the autocorrect feature from URL input box

rocky-singh:
I have added Qt.ImhNoPredictiveText to the inputMethodHint in AddressBar.qml.

### DIFF
--- a/src/app/webbrowser/AddressBar.qml
+++ b/src/app/webbrowser/AddressBar.qml
@@ -242,7 +242,7 @@ FocusScope {
 
         font.pixelSize: FontUtils.sizeToPixels("small")
         color: addressbar.fgColor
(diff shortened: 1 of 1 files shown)


UniversalSuperBox (approved):
Wow, that was fast!

screenshot20181016_100809451

I've tested this on the Nexus 5, Fairphone 2, and Nexus 7:

* Autocorrect does not appear in the address bar
* The keyboard which only allows URL characters appears
* Webapps do not seem to be affected (that's correct, they don't have this address bar)

Flohack74:
Please try to link to a corresponding issue, if it is existing: https://github.com/ubports/morph-browser/issues/56 - it makes our life much easier.

PR #74, opened 2018-10-25
Scrollbar theme

mateosalta:
This introduces a theme to the scrollbar, marius suggested css that should have a nicely styled scroll bar. The scrollbar has a transparent background, and looks similar to the system scrollbar https://t.me/ubports/165236

Possible regression, some sites may reserve space for the scrollbar on the right side of the screen - resulting in a gap, instead of a nice overlay.

Testing on a few sites to see if the result is sutible suggested.


tjrhodes:
Hey, Looks good on white backgrounds, gets lost on dark backgrounds. Rather than being an overlay it appears to take up page width?

andrisk-dev:
nexus 5 2018-w48 - basically the same observations as tjrhodes posted above

UniversalSuperBox:
Yes, it looks like the scrollbar does take up some width on the page. This is expected compared to most desktop browsers, which also have the scrollbar taking up page width (about 17 pixels, I found by cursory searching). The scrollbar being visible on dark backgrounds is not expected, though. @tjrhodes, could you file another bug with your findings?

mateosalta:
yeah, ubuntu.web creates a nice scrollbar. wonder if that code is css? or something diffrent

PR #75, opened 2018-10-26
enable find in page (search text)

balcy:
some context menu items (share, save, ...) not available for webapps
correct navigation requests for webapps (e.g. new view requested)
fixes for "select all" and copy link text action

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -41,6 +41,9 @@ WebView {
     // remember certificate errors for each domain (for click on symbol in address bar)
     readonly property var certificateErrorsMap: ({})
 
+    // better way to detect that, or move context menu items only available for the browser to other files ?
+    readonly property bool isWebApp: (typeof browserTab === 'undefined')
+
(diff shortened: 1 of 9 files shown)


balcy:
up to now all onNewViewRequested for webapps were not working (they just were open in new tab requests for the browser)
I'm not sure how we should handle them in WebViewImplOxide.qml (webapps)
Should we block non-user initiated ones (popups) or open them always in the overlay ?

balcy:
I think it might be good to open popups from the same domain in the overlay

balcy:
@Xray1970 on the first start of sailbook I've seen that morph was opened with static.facebook.com links, could those be missing in the allowed list ?

ghost:
> @Xray1970 on the first start of sailbook I've seen that morph was opened with static.facebook.com links, could those be missing in the allowed list ?

Hi, i need to check this. thanks for the info! I have check it but it seems the are not missing.

mariogrip (approved):
Awesome! LGTM :+1:

PR #84, opened 2018-10-31
for now restore the old behavior for webapp-container

balcy:
allow resource requests even if not defined in the url patterns

- opening ignored navigation requests in morph only makes sense if the are in the main frame (isMainFrame=true in https://doc.qt.io/qt-5/qml-qtwebengine-webenginenavigationrequest.html#isMainFrame-prop) so that clearly has to be corrected
- instead of that we could either ignore them entirely (requires defining them in the URL patterns or other app specific settings), or accept them

Ignoring them would add tracking protection for the webapps, but the browser itself would still load the requests. On the other hand it could break webapps that rely on the page being fully loaded including 3rd party URLs. For the user it would be so that the same page works in morph, but not in the webapp.

### DIFF
--- a/src/app/webcontainer/WebViewImplOxide.qml
+++ b/src/app/webcontainer/WebViewImplOxide.qml
@@ -246,8 +246,18 @@ WebappWebview {
             return
         }
 
-        if (webview.shouldAllowNavigationTo(url))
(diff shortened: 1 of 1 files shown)


mariogrip (approved):
LGTM :+1:

Issue #8, opened 2018-06-04
Restore previous session - does not work 


doniks:
Repro:
go to settings
confirm that "Restore previous session at startup" is ticked 
open a couple tabs
swipe in from the right
swipe the browser down to close it
open the browser again 

Expected:
the tabs from previous session are restored 

Actual:
Only one tab with search engine opens

tatonuta:
The same problem on:

    Device: Krillin - Bq Aquaris E4.5
    Channel: 16.04 RC
    Build: 2018-W27

ghost:
device:Nexus 5
channel: RC 2018-W48
issue solved. Unable to reproduce, tested with up to 7 tabs open.

UniversalSuperBox:
Session restore works on my Nexus 7 (flo), Nexus 5, and Fairphone 2.

MEstner:
Fixed in 2018-w48 in m10 FHD
Unable to reproduce

PR #105, opened 2018-11-19
Fix session restore

UniversalSuperBox:
Put session restore functionality back into morph-browser.

Closes https://github.com/ubports/morph-browser/issues/8


mariogrip (approved):
LGTM

Issue #99, opened 2018-11-13
All tabs but current lose context


vandys:
On dev, updated to 703, and Morph has returned to an early behavior where it maintains browser
context for only the current tab.  I can have two tabs, but I can clearly see it reload the content
afresh when I switch to the other tab.  This is on a freshly booted phone, nothing else competing for memory.

balcy:
yes before https://github.com/ubports/morph-browser/pull/96 no tabs could be unloaded (even if they were closed)
On my phone (Aquaris 4.5) even with only one tab and no other apps open, I see `qml: System low on memory, but unable to pick a tab to unload` the whole time (and background tabs are always unloaded as you describe)
The calculation of free memory in morph-browser.qml might be inaccurate
`
var freeMemRatio = (MemInfo.total > 0) ? (MemInfo.free / MemInfo.total) : 1.0
 // Under that threshold, available memory is considered "low", and the
 // browser is going to try and free up memory from unused tabs. This
 // value was chosen empirically, it is subject to change to better
 // reflect what a system under memory pressure might look like.
 var lowOnMemory = (freeMemRatio < 0.2)
`
-> the ratio of 20% is the threshold
Up to now it always tried to unload tabs too, but because of an error (see mentioned issue / PR) the unload could not finish.

balcy:
should we change the value to 0.15 or 0.1 perhaps ?

vandys:
You may want to look at "avail" as distinct from "free".  A default Linux kernel tries to use would-be
idle memory to cache content, but that shouldn't make the browser leap to low memory behavior.
It isn't the end of the world to have Linux sweep through its less used pages and reclaim memory, and certainly preferable
to having a browser which doesn't support background tabs!

UniversalSuperBox:
Looks like it's a misnomer: https://github.com/ubports/morph-browser/blob/65859e538d52281cc04c5a6554c403038b0f2389/src/app/meminfo.cpp#L124

`MemInfo.free` is actually the available RAM in the system.

tjrhodes:
Device: FP2
System Build: 2018-W48

With 2 or more tabs open the background tabs are always unloaded for me. Only the active tab retains state.

balcy:
yes, the same for me on BQ Aquaris 4.5, maybe we should still lower the threshold value ?

vandys:
> maybe we should still lower the threshold value ?

You should check "free -h" from a shell, and if "avail" is a moderately large number, then you
[...]

PR #106, opened 2018-11-19
Variablize lowMemoryThreshold to keep tabs open

UniversalSuperBox:
I've noticed that in morph-browser, the out-of-memory tab loader was a
bit excessive. For some reason, it would unload tabs even when the
system had lots of memory available -- 1GB or more!
Turns out, the `freeMemRatio < 0.2` test always returned true, no
matter what. I'm not sure why this happens, probably a QML+JS quirk.
By moving the low memory threshold into a new variable, the test
returns correctly based on the amount of available memory in the system.
No more unloading of every tab!

Fixes ubports/morph-browser#99


balcy:
`qml: System low on memory, but unable to pick a tab to unload` messages are gone, and the background tab was not unloaded (BQ Aquaris 4.5). Good job!

balcy (approved):

Issue #92, opened 2018-11-08
Can't remove all text from search box. 


davidmattson1:
BUG-REPORT

Device: Nexus 5
Channel: Stable
Build: OTA-5
Steps to reproduce
Open Morph Browser
Go to Google.com
Search for 'Test' and search
Try to remove all the text from the search box to search again. (Backspace 4 times)
It shows Test in the search box again
Expected behavior
Allow me to remove all the text and leave it blank

Actual behavior
Deleting all the text makes it show Test again

Logfiles and additional information
Not sure that I can.

Note:  I original opened this under the main project but I'm guessing this should go here.

balcy:
it certainly has to to with the input method.
There is currently one workaround: If you turn off "Spell checking" and "Word suggestions" in System Settings -> Language and Text, this does not happen.
The input method seems not properly being reset when the last letter is deleted.
Another workaround: before removing the last letter, click into the field again twice, so that the letter is no longer underlined, and then delete it.
I'm not sure if we can solve this in morph properly, or if it is a problem of QtWebengine itself.
Maybe we can call Qt.inputMethod.reset() when the keyboard opens to get rid of the problem

ruedigerkupper:
Confirmed.

balcy:
the problem is even a bit worse: If you are in a textarea, and press `abc<Enter>` and then `<Backspace>` you end up with `aab`, if you delete that you get `abc` again, which will reappear if you delete it.

And I've seen cases that allow the deletion of the whole text, but one more press on backspace will make the last word reappear. Actually we had some of those problems with Oxide too...

There are more cases like this, so it would not be enough to reset the input method when you enter a field.

balcy:
for some reason the last remembered word is in some cases cached, where it should not be:
e.g.
for all examples only type the letters, not the < and > symbols!

---
* type `>abc<` to the duckduckgo input field (or any other), it is underlined.
* press backspace 3 times -> the input is empty
* press backspace again (in the empty field) -> nothing happens
-> OK
---
* type `>abc <` (space after the c) to the duckduckgo input field
* press backspace -> abc is underlined again
* press backspace 3 times -> the input is empty 
* press backspace again (in the empty field) -> abc (underlined) appears again
-> NOT OK
---
* type `>abc de <` (spaces after c and e) to the input field
* press backspace (de is underlined)
* press backspace 3 times (the word is now "ade" with "de" underlined)
* press backspace 3 times (the word is now "de" which is underlined)
[...]

PR #110, opened 2018-11-22
temporary input issue fix.

balcy:
the input-method-handler (cpp part) does reset the input method after each completed word.
as long as you stay within one text field or different text fields without closing the on screen keyboard, this workaround worked for me well on BQ Aquaris 4.5
the QML part (WebViewImpl.qml) is only needed for entering text fields without the on screen keyboard being open, or if the keyboard type does change
it unfortunately does require a timer, that resets the input method with a delay of 500ms.
altogether the goal here is to improve the input behavior described in https://github.com/ubports/morph-browser/issues/92 as much as possible, while the root cause has not been solved.

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -20,6 +20,7 @@ set(COMMONLIB webbrowser-common)
 set(COMMONLIB_SRC
     browserapplication.cpp
     favicon-fetcher.cpp
+    input-method-handler.cpp
     meminfo.cpp
     mime-database.cpp
     session-storage.cpp
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -664,55 +664,35 @@ WebView {
     onFullScreenRequested: function(request) {
        request.accept();
    }
-/*
-    onDownloadRequested: {
-        if (!request.suggestedFilename && request.mimeType &&
-            internal.downloadMimeTypesBlacklist.indexOf(request.mimeType) > -1) {
-            return
-        }
 
-        if (downloadLoader.status == Loader.Ready) {
-            var headers = { }
-            if (request.cookies.length > 0) {
-                headers["Cookie"] = request.cookies.join(";")
-            }
-            if (request.referrer) {
-                headers["Referer"] = request.referrer
-            }
-            headers["User-Agent"] = webview.context.userAgent
-            // Work around https://launchpad.net/bugs/1487090 by guessing the mime type
-            // from the suggested filename or URL if oxide hasn’t provided one, or if
-            // the server has provided the generic application/octet-stream mime type.
-            var mimeType = request.mimeType
-            if (!mimeType || mimeType == "application/octet-stream") {
-                mimeType = MimeDatabase.filenameToMimeType(request.suggestedFilename)
-            }
-            if (!mimeType) {
-                var scheme = request.url.toString().split('://').shift().toLowerCase()
-                var filename = request.url.toString().split('/').pop().split('?').shift()
-                if ((scheme == "file") || (filename.indexOf('.') > -1)) {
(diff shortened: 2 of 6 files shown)


balcy:
... If you rather don't want to merge it, that's fine.
But it would be interesting to see if it works on other phones like on BQ Aquaris 4.5

UniversalSuperBox:
This does seem to break a fairly important interaction: backspacing after an autocorrect to undo the correction. 

However, the problem behavior that this commit solves appears in other webapps and is fixed by this commit. I'm not sure which option is worse in this case.

balcy:
I'd say if you actually want to write a message with more than 1 line, the behavior without the PR is worse:
for example write
`>first second`
`third <`  (blank after the third)
then one (!) backspace will change that to:
`first ssecondthird` (with secondthird underlined)
pressing backspace many times will end up with an underlined `secondthird` that you cannot delete because it reappears when you delete the s.

Another thing: I've stumbled upon a commit from 4 days ago in QtWebEngine 5.12:

https://github.com/qt/qtwebengine/commit/44abfea789b8677a741530c44a0cd35e5276ddf3

the added test case contains `// 5. Mimic behavior of QtVirtualKeyboard with enabled text prediction.` so it might have to do sth. with our bug.

UniversalSuperBox (dismissed):
After a discussion with @cibersheep, we think it is better to merge this temporary fix and deal with the autocorrect problem. Merge when ready, @balcy.

balcy:
@UniversalSuperBox  is it okay to add search engine "DuckDuckgo (Html)" with this PR as well ?
The normal one has one negative behavior: If the input field has focus, the focus is lost with another click (e.g. other word), and it is quite useful for low spec devices I think.

UniversalSuperBox:
Maybe the name isn't quite right... DuckDuckGo (HTML-only) might make more sense... Either way I'd prefer it in two PRs so that it's easier to revert when the time comes.

UniversalSuperBox (approved):

UniversalSuperBox:
Note to QA: Please use the browser normally to test this issue. Report any and all strange occurrences when typing text into input fields.

UniversalSuperBox:
We've received feedback that the browser now resets the keyboard to the standard (letters) page every time a character is entered from an alternative page. For example, by pressing the `?123` key and then typing a number. Entering text in the URL bar does not have the same problem. Unfortunately, it seems like this problem is related to this patch. We may need to revert it.

cibersheep:
We knew this could happen :)

balcy:
It's a pity... For normal input fields I could have made it so, that only if the last char of the commit string is a blank, the reset is done. But like that I could not detect when the return key was pressed in textareas.
~~for the numbers I've got a solution here:~~
to skip numbers and special chars when doing the input method reset, I've made the following change:
https://github.com/ubports/morph-browser/pull/114
~~but that would have to be extended to the other "special" chars.~~

balcy:
... actually there are already cases in that the keyboard itself switches the layout: e.g. the apostrophe / single quote

tjrhodes:
Lots of issues here still. The one with not being able to delete the "final" word of a phrase as the autocorrect would jump in and restore it seems to be fixed. 

However, return in a multiline input where return would take you to a new line causes the soft keyboard to close. 

The soft keyboard switches to the numbers/symbols input on it's own when you retype in the field and the soft keyboard comes back.

Deleting with backspace gives no autocorrect till you type and then only on the last inputted portion of a word.

(On Fairphone 2)

UniversalSuperBox:
@tjrhodes, I'm not able to reproduce the issue with multiline inputs... could you link a page where the issue occurs, and give exact steps (the buttons you pushed) to reproduce the second?

The third issue is expected with this fix, it commits the word into the "completed text" buffer, so to speak, so backspacing doesn't start autocorrect again.

PR #114, opened 2018-11-25
input-method-handler: only reset the input method, if last char in the commit string is a letter, space, or single quote

balcy:
when entering a number at the end of a word, it is automatically commited. (no reset needed)

### DIFF
--- a/src/app/input-method-handler.cpp
+++ b/src/app/input-method-handler.cpp
@@ -26,7 +26,14 @@ bool inputMethodHandler::eventFilter(QObject* obj, QEvent* event)
     QInputMethodEvent * inputevent = static_cast<QInputMethodEvent*>(event);
 
     //qDebug() << "input event, predit string: " << inputevent->preeditString() << " commit string: " << inputevent->commitString();
-    // check the request contains informatino about the text format
+
+    // if the last char of the commit string is a number, do nothing (reset would change the keyboard layout back to letter mode)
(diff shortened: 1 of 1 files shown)


balcy:
should be safe to apply, because it does the same action as before, but in fewer situations.

UniversalSuperBox:
i'm still getting the same behavior on the Nexus 5 and Nexus 7. The keyboard resets to letters immediately after typing a number or special character.

balcy:
on my phone I could not install it via the QA script, but installed the deb packages directly. Did the install process work on your phone ? I had to downgrade even, maybe because of the other  change in xenial ? Strange that it works on my phone..

balcy:
can you retry with installing the artifacts
morph-browser_0.24+ubports+0~20181125112809.2~1.gbpb7c942_armhf.deb
qml-module-morph-web_0.24+ubports+0~20181125112809.2~1.gbpb7c942_armhf.deb
from the Jenkins result ? Or alternatively with the click package of this branch ?

UniversalSuperBox:
`sudo apt install morph-browser=0.24+ubports+0~20181125112809.2~1.gbpb7c942 qml-module-morph-web=0.24+ubports+0~20181125112809.2~1.gbpb7c942` got me the right version. Works as expected.

UniversalSuperBox (approved):

PR #121, opened 2018-12-02
display error page if page load fails

### LINKED ISSUE #118: Broken link without error
Device: m10 HD Xenial RC
Steps: go to a inexistent url (for example: http://www.aceadventuregames.com/ this link desn't exist anymore)
Behaviour: weird blank page with three buttons?
Expected: Error message saying «Link couldn't be opened» or similar

**Images**

Morph:
imatge

Oxide:
imatge

(«Network error. It seems that you are having problems to show _link_. Check your connection and try to load the page again»)

### LINKED ISSUE #133: Initial page of webapps cannot be loaded sometimes
In the current xenial branch, but in the PR branch xenial_-_loadfailed, too, the following problem exists:
Steps:
- open a webapp
- close it after it is fully loaded
- start it again after some seconds
Problem:
- the error sheet (Network error) appears
- when tapping refresh the webapp loads successfully
For me this problem is more likely to happen on Wifi connection
So far I've not seen the problem for local html apps.

I think those at not network errors but rather the problem of the webapp container startup, maybe a race condition.

balcy:
ErrorSheet: in Browser.qml use the property lastLoadFailed for the binding (was already correct in WebApp.qml)

Browser.qml / WebApp.qml: remove the top margin of the error sheet
AddressBar: new lastLoadSucceeded property: show lock symbol only, if last load was successful for https pages

fixes https://github.com/ubports/morph-browser/issues/118

now the chances are high that it fixes https://github.com/ubports/morph-browser/issues/133 as well

### DIFF
--- a/src/Morph/Web/MorphWebView.qml
+++ b/src/Morph/Web/MorphWebView.qml
@@ -320,7 +320,7 @@ WebEngineView {
     readonly property bool lastLoadStopped: false // TODO internal.lastLoadRequestStatus === Oxide.LoadEvent.TypeStopped
     readonly property bool lastLoadFailed: internal.lastLoadRequestStatus === WebEngineLoadRequest.LoadFailedStatus
     onLoadingChanged: {
-        if (loadRequest.status != WebEngineLoadRequest.LoadStartedStatus) {
+        if ((loadRequest.url === url) && (loadRequest.status !== WebEngineLoadRequest.LoadStartedStatus)) {
             internal.lastLoadRequestStatus = loadRequest.status
         }
         internal.dismissCurrentContextualMenu()
--- a/src/Morph/Web/MorphWebView.qml
+++ b/src/Morph/Web/MorphWebView.qml
@@ -303,6 +303,7 @@ WebEngineView {
     QtObject {
         id: internal
         property int lastLoadRequestStatus: -1
+        property string lastLoadRequestErrorString: ""
         property QtObject contextModel: null
 
         function dismissCurrentContextualMenu() {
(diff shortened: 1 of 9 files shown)


balcy:
... please do not yet merge, my test have shown that sometimes the initial page shows the 'network error'

balcy:
@UniversalSuperBox  ok I think we can try the current state, as always further testing is appreciated !

balcy:
while https://github.com/ubports/morph-browser/issues/133 is not solved by that, at least it does not make it worse..

UniversalSuperBox:
I've found an issue with the display of this page, or lack thereof, when the user does something strange...

If I do the following, I get the incorrect error page shown before:

1. Navigate to a page which causes a page error, such as http://asdf.fdsa.asdf
1. Navigate to a correctly working page, such as ddg.gg
1. Navigate to another working page, such as ubports.com
1. Press the back button twice in quick succession

This only occurs if the problem page is the first page in a back/forward history, not if it's in the middle of several working pages.

balcy:
Since I'm without computer these days, I suggest to leave it out for now. (or even better s.o. else finds the solution :)  )
If I switch back and forth between the not existing page and ubports.com, I get the problem too (no error notice) sporadically.
At least I know now that ddg.gg does exist ;)

balcy:
what i still not get is why the error pages provided by QtWebEngine / chromium look so weird (2 small rectangles and 1 bigger rectangle / button with a download symbol)
If you save the source code of them, it looks like an email sent by "Blink".
If they were proper error pages, we would not need our ErrorSheet here.
there is even a property to turn the error pages off in WebEngineSettings, but then the browser does not navigate to that page at all. (http://doc.qt.io/qt-5/qml-qtwebengine-webenginesettings.html#errorPageEnabled-prop) Maybe we than can use a dialog instead.
Will look at it 2019 i guess, suggestions welcome :)

balcy:
does the problem still exist with the latest commit ? (I could no longer reproduce it)

UniversalSuperBox:
I'm not able to reproduce that behavior either. I'm still sometimes able to get the incorrect error page, but not in any reproducible way (and admittedly, only by pressing buttons in series that would not make sense for any usage). Good work!

UniversalSuperBox (approved):

PR #128, opened 2018-12-22
Add theming support...

nfsprodriver:
... to keep everything readable after changing the UI theme. Feel free to optimize it!
You can switch your theme best way by using UT Tweak tool latest version.

### DIFF
--- a/src/app/ChromeBase.qml
+++ b/src/app/ChromeBase.qml
@@ -55,7 +55,7 @@ StyledItem {
         id: backgroundRect
 
         anchors.fill: parent
-        color: Theme.palette.normal.background
+        color: theme.palette.normal.background
 
         Rectangle {
             anchors {
@@ -64,7 +64,7 @@ StyledItem {
                 bottom: parent.bottom
             }
             height: units.dp(1)
-            color: "#dedede"
+            color: theme.palette.normal.background
         }
     }
 
--- a/src/app/ChromeBase.qml
+++ b/src/app/ChromeBase.qml
@@ -64,7 +64,7 @@ StyledItem {
                 bottom: parent.bottom
             }
             height: units.dp(1)
-            color: theme.palette.normal.background
+            color: theme.palette.normal.base
         }
     }
 
--- a/src/app/ChromeButton.qml
+++ b/src/app/ChromeButton.qml
@@ -26,7 +26,7 @@ AbstractButton {
 
     Rectangle {
         anchors.fill: parent
-        color: Theme.palette.selected.background
+        color: theme.palette.selected.background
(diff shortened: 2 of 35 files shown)


myii:
@nfsprodriver I've only started testing this but the improvement is undeniable.  The `SuruDark` experience is so much better.  Thanks for working on this.

nfsprodriver:
Thank you @myii ! Well, if there's nothing to add/change from your side, I'd like to remove the WIP status and it's ready to get merged :)

balcy:
thank you for adding the theme support, the only small change I would do (I'm using the default theme) is to use slightly other colors for the light grey. (e.g. bottom bar), to match the previous colors when possible.
First I thought maybe Qt.lighter() / Qt.darker() can be used, but maybe it is better to find equivalents in the theme.
(e.g. theme.normal.foreground / theme.normal.overlay for the #f6f6f6)
I try sending some small change tomorrow.

nfsprodriver:
Okay, good point. As I did that work, I didn't have the color palette in range. But I think I'll have a look at the UI theme definitions. We may add a table of the color palette for both Ambiance and Suru to the docs.

balcy:
looks fine now, just have found 2 little things (color of the menu / tab count) for which I've made a PR to this branch. 👍

balcy (dismissed):

myii:
@nfsprodriver @balcy Thanks for the ongoing work.  You're probably already aware of this; the tab count is now being displayed but with a strange outline.

Using the latest commit:

```bash
phablet@ubuntu-phablet:~$ dpkg -l | grep morph-browser
ii  morph-browser                                        0.24+ubports+0~20181230145317.5~1.gbp63ced9                  armhf        morph web browser
```

We've got:

screenshot20181230_171150211

nfsprodriver:
Hm, on my N5 everything looks fine for both themes.

myii:
@balcy Excellent, that's fixed now, thanks.

screenshot20181231_002525507

balcy (approved):

balcy:
oops forgot to squash the PR (phone screen too small), hopefully next time.

myii:
@nfsprodriver @balcy Spotted a little problem.  Don't think it's worth opening a new issue about, informing you both seems enough.

#### Switching tabs in portrait orientation

`Ambiance`:

screenshot20190102_144642407

`SuruDark`:

screenshot20190102_144045883

balcy:
yes those are mentioned here: https://github.com/ubports/morph-browser/pull/134
Thank you for reporting it.
Also the "SuruGradient" theme has more problems (e.g url in address bar, contrasts in general).

PR #132, opened 2018-12-26
"keep display on" for audio playback and fullscreen (video)

### LINKED ISSUE #71: Standby mode while video playback
When I play videos on the current browser, the playback stops after the defined standby timer (default 1 min)

balcy:
turn off screensaver while the browser window is active and "recentlyAudible" or "isFullScreen" property of the current webview is true

fixes https://github.com/ubports/morph-browser/issues/71

### DIFF
--- a/src/app/BrowserWindow.qml
+++ b/src/app/BrowserWindow.qml
@@ -18,6 +18,7 @@
 
 import QtQuick 2.4
 import QtQuick.Window 2.2
+import QtSystemInfo 5.0
 import Ubuntu.Components 1.3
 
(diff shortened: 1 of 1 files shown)


balcy:
do you think we should create a setting entry for that ? Some users might not know, that the phone stays unlocked in those situations.

mariogrip:
Awesome!

This was the default behavior of oxide, so i think its fine for now :)

But an option would be a good thing to add tough.

mariogrip (approved):
LGTM :+1:

PR #134, opened 2018-12-30
2 minor changes

balcy:
- adapt background color of the menu (to the previous one)
- tab count was not visible in SuruDark (background is an image here)


nfsprodriver (approved):
That looks fine, thank you! We may find a solution for the images used in some parts like in the tab list headers.

balcy:
yes maybe those can be changed, too

Issue #124, opened 2018-12-02
Cannot inject custom js in webapps


mymike00:
I have a plain webapp (no qml at all, I have `Exec=webapp-container` in the `.desktop` file) and a file called `webapp-properties.json` which calls a custom script.
the code in this custom script was executed with oxide, but not anymore with morph

cibersheep:
There's a workaround and is using a qml with a webview and embed the js as a UserScript
In the import area you could:
```import Morph.Web 0.1
import QtWebEngine 1.7 as Oxide
```

And in a Page element
```
WebContext {
	id: webcontextIF

        userScripts: [
            Oxide.WebEngineScript {
                id: cssinjection
                injectionPoint: WebEngineScript.DocumentCreation
                sourceUrl: Qt.resolvedUrl("js/cssinjection.js")
                worldId: WebEngineScript.MainWorld
            }
	]
}
```

And the webview:
```
WebView {
	id: webview

	context: webcontextIF
	url: 'https://whatever'
}
```

mymike00:
Thanks, I'll try that workaround, but I highly prefer not using qml in simple webapp like mine...

balcy:
do you have an example of an app using it ? I've found hints to webapp-properties.json in the source code of morph, but not understood how the properties really work.

mymike00:
my `.desktop` has
```
Exec=webapp-container --webappModelSearchPath=. --store-session-cookies --webappUrlPatterns=https?://www.mywebsite.com/* https://www.mywebsite.com/mypath %u
```
`.apparmor` is
```
{
    "template": "ubuntu-webapp",
    "policy_groups": [
        "keep-display-on",
        "networking",
        "webview"
    ],
    "policy_version": 1.3
}
```
then I have a file called `webapp-properties.json`
```
{
    "includes": ["https://www.mywebsite.com:*/*"],
    "name": "ExtendedWebappProperties",
    "scripts": ["myscript.js"],
    "domain":"",
    "homepage":""
}
```
and obviously a script called `myscript.js`

UniversalSuperBox:
Closed via #140

Fuseteam:
This was appearently broken with the removal of oxide

balcy:
[...]

PR #140, opened 2019-01-12
add custom user scripts for webapps defined in webapp-properties.json

balcy:
solves https://github.com/ubports/morph-browser/issues/124

### DIFF
--- a/src/app/webcontainer/webapp-container.qml
+++ b/src/app/webcontainer/webapp-container.qml
@@ -182,6 +182,40 @@ BrowserWindow {
             var context = item.currentWebview.context
             onlineAccountsController.setupWebcontextForAccount(context)
             item.currentWebview.settings.localContentCanAccessRemoteUrls = localContentCanAccessRemoteUrls
+
+            loadCustomUserScripts()
+        }
+
+        function loadCustomUserScripts() {
+
+            var scripts = [];
+
+            // app specific user scripts
+            var idx = webappModel.getWebappIndex(getWebappName())
+            var customScripts = webappModel.data(idx, UnityWebApps.UnityWebappsAppModel.Scripts)
+
+            if (customScripts.length === 0)
+            {
+                return;
+            }
+
+            var i;
+            for (i = 0; i < customScripts.length; i++)
+            {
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
I notice that some lines here have semicolons while others do not. Proper JS style says that all lines should end with semicolons. I understand that the rest of this file does not always follow proper style, but do you think that new contributions should?

balcy:
yes I think it is better with all lines having semicolons, but as far as I remember most of morph / webbrowser app so far was without them.
Yes for new contributions I think we should try that.

UniversalSuperBox:
Alright, this tests out fine to me. I'll wait for you to fix that, test again, then merge.

UniversalSuperBox (approved):

cibersheep:
@balcy Is there a blog / doc / whoeva' somewhere to know how to use this cool feature?

balcy:
I haven't found a good documentation so far, it was new to me, too and I asked the author of the issue https://github.com/ubports/morph-browser/issues/124
The scripts injection is / was a feature "unity webapps" of https://github.com/ubports/unity-webapps-qml/blob/b8a72650d2172b2ada9c217b148772179720e6bf/src/Ubuntu/UnityWebApps/plugin/unity-webapps-app-model.cpp and webbrowser-app made use of them.
They seem not to have been widely used though. (e.g. https://docs.ubuntu.com/phone/en/web/ubuntu-webapps-guide does not mention the user scripts)
Thinking about it now it seems to be quite a legacy feature (used in unity webapps like https://packages.ubuntu.com/xenial/all/unity-webapps-bbcnews), so we could think about additionally make available a more straight-forward way to add user scripts, e.g. via a command line switch like the url patterns and the user agent, we could add "--scripts" to add user scripts.

mymike00:
@balcy as you want, the important thing for me is to be able to run the custom script, I don't have preferences on how it is run.
I also didn't know and didn't find any docs on this topic, @stuartlangridge helped me in the telegram group and he told me to do it in that way...

balcy:
ok alright : ) I think it was good to solve the regression first. At least now new users can use the issue as an example how to use custom scripts.
The only possible problem I've seen so far that url patterns given via command line seem to be treated differently if webapp-properties.json is used.

PR #141, opened 2019-01-12
bring back favicons

balcy:
favicon-fetcher: use the image:// url provided by QtWebEngine (do not cache them with morph)

for bookmarks still use the local cache for favicons

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -221,7 +221,7 @@ BrowserView {
         Actions.Bookmark {
             enabled: currentWebview
             // QtWebEngine icons are provided as e.g. image://favicon/https://duckduckgo.com/favicon.ico
-            onTriggered: internal.addBookmark(currentWebview.url, currentWebview.title, currentWebview.icon.toString().substring(16))
+            onTriggered: internal.addBookmark(currentWebview.url, currentWebview.title, (currentWebview.icon.toString().substring(0, 16) == "image://favicon/") ? currentWebview.icon.toString().substring(16) : currentWebview.icon)
         },
         Actions.NewTab {
             onTriggered: internal.openUrlInNewTab("", true)
@@ -591,7 +591,7 @@ BrowserView {
         onToggleBookmark: {
             if (isCurrentUrlBookmarked()) BookmarksModel.remove(tab.url)
             // QtWebEngine icons are provided as e.g. image://favicon/https://duckduckgo.com/favicon.ico
-            else internal.addBookmark(tab.url, tab.title, tab.icon.toString().substring(16))
+            else internal.addBookmark(tab.url, tab.title, (tab.icon.toString().substring(0,16) == "image://favicon/") ? tab.icon.toString().substring(16) : tab.icon)
         }
         onWebviewChanged: bookmarked = isCurrentUrlBookmarked()
         Connections {
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -221,7 +221,7 @@ BrowserView {
         Actions.Bookmark {
             enabled: currentWebview
             // QtWebEngine icons are provided as e.g. image://favicon/https://duckduckgo.com/favicon.ico
-            onTriggered: internal.addBookmark(currentWebview.url, currentWebview.title, (currentWebview.icon.toString().substring(0, 16) == "image://favicon/") ? currentWebview.icon.toString().substring(16) : currentWebview.icon)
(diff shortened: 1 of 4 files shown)


balcy:
note: 

- newly created bookmarks will cache the icons (like webbrowser-app did)
- existing bookmarks can be "repaired" by removing and adding them again as bookmark. They have (until this PR) saved the "image://favicon/\<FaviconUrl\>" url, which can only used while QtWebEngine has temporarily cached the icon (e.g. tab of this page currently open)

src/app/webbrowser/TabComponent.qml
```diff
@@ -363,12 +364,18 @@ Component {
...
+
                 if (webviewInternal.storedUrl.toString()) {
-                    HistoryModel.update(webviewInternal.storedUrl, webviewInternal.title, icon)
+                    HistoryModel.update(webviewInternal.storedUrl, webviewInternal.title, (icon.toString().substring(0,16) == "image://favicon/") ? icon.toString().substring(16) : icon)
                 }
             }
-
+            /*
             //onGeolocationPermissionRequested: requestGeolocationPermission(request)
 */
```
UniversalSuperBox: Is there any reason for leaving this commented-out code in the file? If not, let's just remove it.

src/app/webbrowser/Browser.qml
```diff
@@ -589,7 +590,8 @@ BrowserView {
         onCloseTabRequested: internal.closeCurrentTab()
         onToggleBookmark: {
             if (isCurrentUrlBookmarked()) BookmarksModel.remove(tab.url)
-            else internal.addBookmark(tab.url, tab.title, tab.icon)
+            // QtWebEngine icons are provided as e.g. image://favicon/https://duckduckgo.com/favicon.ico
+            else internal.addBookmark(tab.url, tab.title, (tab.icon.toString().substring(0,16) == "image://favicon/") ? tab.icon.toString().substring(16) : tab.icon)
```
UniversalSuperBox: I feel like something more elegant could be found here... Is there any reason that `tab.icon` couldn't be changed to be appropriate for this use, not requiring all this substring nonsense?

UniversalSuperBox (changes requested):

src/app/webbrowser/Browser.qml
balcy: The crucial point here is, that QtWebEngine does cache the icons, and while the pages / tabs are open, it is no problem to give the URL provided by QtWebEngine like `image://favicon/https://duckduckgo.com/favicon.ico` to the QML favicon image. (so no extra caching needs to be done by morph, we can directly use the locally saved image)
That is why tab.icon should keep that URL in my opinion.
Only for creating the bookmark (and the history) we need the right part of the URL, so that it will be cached and visible in the bookmarks view without accessing the page.
Doing it with the substring was the most transparent way to me here, of course it is possible to use sth. like "if the scheme or the url is *image* and the host is *favicon*, use its path"

balcy: `( (UrlUtils.extractScheme(tab.icon) === "image") && (UrlUtils.extractHost(tab.icon) === "favicon") ) ? UrlUtils.extractPath(tab.icon) : tab.icon` maybe, but UrlUtils.extractPath has not yet been defined in UrlUtils ... At that point I thought it does not look much better so I ended up with the substring, but I'm fine with a regex / sth else too.

UniversalSuperBox: Completing extractPath and adding `schemeIs(url, expectedScheme)` and `hostIs(url, favicon)` is one way this could be completed which would be more readable. Or, well, forget extractPath and just do schemeIs and hostIs for readability. Would be an improvement.

UniversalSuperBox (approved):

jezek:
FP2 (2019-W09)
tested on 3 sites: [gmail.com](https://gmail.com), [memrise.com](https://memrise.com). [slovniky.juls.savba.sk](http://slovniky.juls.savba.sk)

- in tabs all 3 favicons are displayed corectly.
- in bookmarks the favicons where not shown for allready created bookmarks. After bookmark resave, ~~only the favicon for slovniky.juls.savba.sk is not displayed. The other two~~ favicons are correct.

Verdict: tab favicons are all back, bookmark favicons are ~~partialy~~ back ~~(not working for all sites, for which tab favicons are working)~~

Note; in bookmarks, the first is duckduckgo. the favicon is not shown and can not resave, cause I have no bookmark for ddg.

**EDIT**: As for now all bookmark favicons (except the first duckduckgo) are displayed correctly. Maybe a restart was needed, maybe someting was fixed.

ziggutas:
Nexus 7 flo, 2019-W09. Favicons are back

haeckle:
Works on Nexus 4 with 2019-W09

ziggutas:
BQ E4.5 2019-09
Favicons are back...except for 'homepage' in bookmarks which does not display a favicon in the bookmark list, no matter what address one puts in for the homepage via browser settings.

balcy:
yeah that's right the homepage is without icon. I've looked into the code, and there is no icon put in there. Would be sth. we could add im the future.

Issue #64, opened 2018-10-17
Cannot copy current URL


ruedigerkupper:
There seems to be no way of copying the current URL. I can mark the text (light blue background), but then I find not way of copying. The familiar long-press does not do anything in the URL bar...

Second, I tried the „share link“ option from the burger menu. It shows a selection of apps to share with, but nothing happens when I select one.

balcy:
the second one is solved by now: https://github.com/ubports/morph-browser/issues/51
with my current version the copy does work as well for the URL

balcy:
please reopen if the issue is still present on RC or devel

ghost:
device:Nexus 5
channel: RC 2018-W48
issue solved. Unable to reproduce

andrisk-dev:
device: Nexus 5 
build: 2018-W48
Unable to reproduce - can copy url with long press context menu and also through browser menu - share (copy url to new message in messages app).

Mraanderson:
Device: Nexus 5
Channel: RC 2018-W48
Unable to reproduce. Long press on URL presents Cut/Copy option. Cut worked and paste worked.

MEstner:
In 2018-w48 m10 fhd
Copy do what it should. If i open a new tap and try to paste I can't.
If I am already on a  website and then try to paste the url it works as expected.

ruedigerkupper:
I think this is (mostly) solved:
- the hamburger menu still has no item to copy the URL
- I can mark the current URL with a short tap and then do a long press to get an option menu and copy it. That seems to work, although it is not the combination I have come to expect. (That would be the exact opposite: long-press to mark, then short tap to get an option menu what to do with it, like copy the text.)
I don't know if I am somehow skewed in my expectation or if GUI habits have really changed so much over the years.
To be more specific: Isn't the agreed-on way of interacting with text:
 - the user indicates what peace of text she wants to interact with (this is done on the desktop with a right click, which acceptably translates into a long click on the touch interface)
- after that, the user selects from that menu what to do (this is done on the desktop with a left click, which acceptably translates into a short click on the touch interface)

So the expected (and trained) combination is "long-press-then-tap", which fails. What works is "tap-then-long-press", which feels awkward.
Am I somehow skewed, or has this this been just confused in this context?

balcy:
thank you for the observations and suggestions !
[...]

PR #153, opened 2019-02-21
address bar context menu

balcy:
- new address bar dialog (copy url, paste & go, select all)
- add "copy url" to browser menu, change some positions
- fix empty favicons in the HistoryModel

(see suggestions in https://github.com/ubports/morph-browser/issues/64)


UniversalSuperBox:
I'll need to defer to @cibersheep on this one. I feel like the overflow menu is starting to, erm, overflow.

Menu that appears when you long-press on the unselected address bar:

image

Overflow menu:

image

cibersheep:
We need a plan here. As Dalton pointed, the main menu has become too busy, we have 3 different layout for main menu, copy & paste in the address bar, copy paste in the body and the address bar pop up you present.

Talking about this pop up: I don't think is needed as if you tap on the address bar and log tap on the url you can already copy it, (x) delete it and paste, etc. I think it adds more busyness

balcy:
ok so the initial feedback for this was https://github.com/ubports/morph-browser/issues/64#issuecomment-460846753

The reason for the new dialog (for me) was:
- there was no long press (right click) action for the address bar before
- pasting an url quickly now takes a tap, long press, tap on "paste", tap on enter
  the new menu would only take a long press, and tap on "paste & go"
- the paste & go is available for firefox / other browsers too

does s.o. know if it is possible to make the textbox-context-menu appear on a single long press / right click without selecting the text first ?

Just doing nothing for a long press or doing the same as a normal tap does (select all text) seems wrong, too.

cibersheep:
I misunderstood the use, I thought you added an action to the main menu.
- Check the code of the address bar as it selects the whole text on single tap.
- I wouldn't use a popup for the menu, instead I would add paste&go to the copy/paste menu
imatge when long tapping on the address bar
- I would not add a Copy url on the main menu, just use the above copy/paste menu

balcy:
new version: (long press on unselected address bar)
image

note: the other context menu for the text box (when already selected) is part of the UI toolkit and cannot be easily extended IMHO

UniversalSuperBox (approved):
A bit unfortunate that we can't add things to the general context menu, but that's handled by Maliit and we need to think about how we handle it pretty soon anyway. Something is better than nothing!

I'm sorry about the delay to review this again. Feel free to ping me if something goes this long again

cibersheep:
Recently I've looked into the context menu code. Is in [here](https://github.com/ubports/ubuntu-ui-toolkit/blob/xenial/src/imports/Components/1.3/TextInputPopover.qml#L27)
Adding an Action for every element you want. the `target` is the element you press and hold on to but it will be on every context menu ;)

Danfro:
Testing on E5, RC channel 2019-W18.

Context menu mostly works fine. Two things to mention:

1. "paste and go" seems to work only  on the start-page with _https://duckduckgo.com_. I could not get this option at another point/page.
I did select all on a bookmark before testing with press and hold to raise the context menu. I never got "paste & go" except on the start-page.
Especially after clearing the address via "X" icon you would expect "paste&go" to be available.

2. a) "Select all" is ONLY available if there is just the single cursor (no other selection) AND you press & hold onto that cursor. It would be great if press/hold somewhere else could do that as well and if "select all" would be available also with another selection. Basically have it as standard button available in the context menu?

2. b) After pressing "select all"  the context menu for cut/copy/paste appears. But this is located somewhat to the right hand side of the screen. With me being right handed, that is where my hand was still hovering. It took me three attempts to realise that the context menu appeared because it was hidden underneath my hand. Could the context menu always pop up in the centre of the screen? I am not suggesting on the left hand side because that would just cause the same issue for every on left handed.

UniversalSuperBox:
@Danfro, 

The "Paste and Go" only appears if the address bar is deselected and you press and hold on it in the deselected state. It will not appear if you have already selected text or have any cursor in the field.

Other issues with selection behavior are system-wide and could be filed on https://github.com/ubports/ubuntu-ui-toolkit.

Danfro:
@UniversalSuperBox 
Even if the cursor is within the address bar (without actual selection) there is no "paste and go".
That's what I think is not intuitive. Why only if address bar is entirely deselected? Why does it need this restriction? I do not see the need for this.

UniversalSuperBox:
See the comment above, https://github.com/ubports/morph-browser/pull/153#issuecomment-467997916

Danfro:
OK. Sorry. Did miss that note. ;-)

jezek:
FP2 (2019-W18)

context menu works

ziggutas:
BQ E4.5 rc (2019-W18)
Long press on deselected address bar offers the 'paste and go' option

ziggutas:
Nexus 7 (flo) rc (2019-W18)
context menu works

Issue #154, opened 2019-02-23
save domain specific settings


balcy:
I'd suggest to add another data model (like BookmarksModel), that saves domain specific settings.

For each domain (if not using incognito mode) the following could be saved:

- custom zoom level (if not using the default zoom), see https://github.com/ubports/morph-browser/issues/107
- a flag that indicates if custom URL schemes are allowed (user is prompted when a page wants to use a custom url scheme), see https://github.com/ubports/morph-browser/issues/147
- ~~possibly a custom user agent string and / or desktop mode flag, see https://github.com/ubports/morph-browser/issues/109#issuecomment-441133462~~
- permissions (e.g. always allow location / audio for this page)
- ...

This should not be done for all domains / visited pages, but only for each domain with settings other than the default.

The settings should be cleared when the user deletes the browser history.
Possibly in the settings menu single entries could be removed / edited, too.

Flohack74:
Would that be also possible for WebApp containers? I have a container for public transport and I have to give permission for location every time again, its annoying ;)

balcy:
I think so. If the webapp container can access the database and we assume the setting is valid for all webapps and morph-browser.

PR #159, opened 2019-02-24
- fix empty favicons in the HistoryModel

balcy:
- error page only is cleared if document element exists


UniversalSuperBox:
Does this fix a specific issue or behavior?

balcy:
yes for the recent "favicon" PR the icons for the HistoryModel have not always been cached correctly. (The database contained empty entries)
with this patch it was reliable for me.

UniversalSuperBox:
How about the documentElement change, then? Does this fix the flickering error page?

balcy:
no it doesn't, but for any error page that does not have a document element, the java script command cannot work.
For the flickering I rather plan to allow custom schemes generally in QtWebEngine, but ask the user when a custom URL is used for the first time for a domain (which has to be confirmed by the user, see https://github.com/ubports/morph-browser/issues/154)

UniversalSuperBox (approved):

PR #174, opened 2019-03-30
Xenial   errordarktheme

### LINKED ISSUE #164: Using SuruDark theme, error texts are not visible
Device: FP2 (2019-W09)

Steps to reproduce:
- Using [UT Tweak Tool](https://open-store.io/app/ut-tweak-tool.sverzegnassi) select SuruDark theme & restart Unity8
- Open new, non-existent page in morph browser (eg. www.ubports1234.com)

Expexted behaviour:
- error message is show, like picture in https://github.com/ubports/morph-browser/issues/118#issuecomment-468712686

Current behaviour:
- error message text is not visible, like picture in https://github.com/ubports/morph-browser/issues/118#issuecomment-468719770

balcy:
fixes https://github.com/ubports/morph-browser/issues/164

### DIFF
--- a/src/app/BrowserView.qml
+++ b/src/app/BrowserView.qml
@@ -42,6 +42,11 @@ FocusScope {
     }
     property alias actions: unityActionManager.actions
 
+    Rectangle {
+      anchors.fill: parent
+      color: theme.palette.normal.background
+    }
+    
     default property alias contents: contentsItem.data
     property alias automaticOrientation: contentsItem.automaticOrientation
     OrientationHelper {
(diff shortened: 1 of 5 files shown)


UniversalSuperBox:
When you say "Text files will no longer be readable with dark theme", how would I test that? I tried opening up a build log from ci.ubports.com (which should be a text file) and it appeared fairly normally.

balcy:
Hi Dalton, actually that comit only took back another commit with that I have set the background color of the webview, to the background color of the theme.
I reverted it because .txt files would then be black on black, because the default foreground color for the webview cannot be set via a property, and stays black even with the dark theme.

UniversalSuperBox:
Ah, that explains why it works now!

UniversalSuperBox (approved):

Issue #167, opened 2019-03-07
Context dialog in Incognito window opens in the main window


kugiigi:
## Steps to reproduce
1. Open Morph browser
2. Open an Incognito window/tab
3. Load a page.
4. Press and hold a link in the page

## Expected Result
A context menu is displayed showing options such as "Open in new tab" and "Open in new window"

## Actual Result
The incognito window doesn't show anything as if it's not reacting to the press and hold gesture but if you switch to the main non-incognito window, you'll see that the context dialog is displayed there instead.

balcy:
yes confirmed.
Even requests, if the page (in private mode) can access the microphone, do appear on the not private page.

njtwiseman:
Confirmed I see it

jezek:
FP2 (2019-W18)
Cotext menu is displayed properly in incognito window.

ziggutas:
BQ E4.5 rc (2019-W18)
Context menu displays ok

ziggutas:
Nexus 7 (flo) rc (2019-W18)
Context menu displays ok

Issue #116, opened 2018-11-28
"Back to Safety" on certificate error dialog does not go back


UniversalSuperBox:
When browsing to a site with an SSL certificate error (try any of those on https://badssl.com), the certificate error dialog appears as expected:

![image](https://user-images.githubusercontent.com/21966173/49159282-bc954100-f2e9-11e8-9adf-84282d9977d2.png)

I tap "Back to safety".

### Expected behavior

The browser cancels navigation, then goes back to the site I was previously viewing.

### Actual behavior

The browser is left on a blank page. I must tap the Back button to continue browsing from where I left off.

UniversalSuperBox:
Since I'm marking this as a "good first issue", I'm also going to offer some helpful hints and places to look in the code. Only a cursory understanding of QML is required to fix this issue, and you can ask in the Matrix room #ub_bootcamp:matrix.org for help.

## Hints to fix

The dialog is defined in [CertificateVerificationDialog.qml](https://github.com/ubports/morph-browser/blob/fb0acef99af01b72b428f13280d02bdf2de2ea6d/src/app/CertificateVerificationDialog.qml). There are two signals, `accept()` and `reject()`, which are sent from this dialog. In order to actually use a signal, you need to *connect* it to another signal or QML function. When a signal is connected to a function, the function is run when the signal is sent.

Actually showing the dialog on a certificate error is performed at: https://github.com/ubports/morph-browser/blob/1540ddc7664f7ecd818aa550c99105944d85e08e/src/app/WebViewImpl.qml#L217

A good solution will instruct `webview` to go back after *rejecting* the certificate error.

balcy:
it might be good to take the canGoBack (http://doc.qt.io/qt-5/qml-qtwebview-webview.html#canGoBack-prop) property into account, because the insecure site could have been defined as homepage. (or being restored in session restore)

There might be one edge case, that could cause problems: If the previous page (with valid certificate) for some reason redirects to the broken page automatically, the user might end up in a loop where they can only close the browser or accept the navigation to the insecure page.

balcy:
with https://github.com/ubports/morph-browser/pull/121 instead of a blank page an error page is shown

balcy:
maybe we can add a Cancel button to the dialog, that just closes it (as the go back to safety button now does)
The back to safety button (if needed at all) should then only be visible if going back is possible.
[...]

PR #176, opened 2019-04-07
added function returnIfPossible, to leave certificate error

njtwiseman:
Somewhat better:

Added a function "returnIfPossible" which

A. Checks to see if the browser is able to go back
B. Goes back if the above returns true.

Problem is that it seems like it goes back twice now.  Just wanted to submit a little halfway point work since this is the first time I'm contributing

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -248,7 +248,12 @@ WebView {
          }
     }
 
-    function returnIfPossible(){
+    // returnIfPossible -> rejects bad certificate and returns to last page
+    function returnIfPossible(certificateError){
+        // deferred rejection so that c.v.Dialog isn't called twice
+        certificateError.rejectCertificate
+
+        // check if back is available
         if (currentWebview.canGoBack) 
         {
             currentWebview.goBack()
@@ -264,8 +269,8 @@ WebView {
         certificateVerificationDialog.localizedErrorMessage = certificateError.description
         certificateVerificationDialog.errorIsOverridable = certificateError.overridable
         certificateVerificationDialog.accept.connect(certificateError.ignoreCertificateError)
-        certificateVerificationDialog.reject.connect(certificateError.rejectCertificate)
         certificateVerificationDialog.reject.connect(returnIfPossible)
+        
         
     }
 
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -307,7 +307,7 @@ WebView {
                     quickMenu.visible = true;
                 }
 
-                var commandGetContextMenuInfo = ""
+                var commandGetContextMenuInfo = "
                 var morphElemContextMenu = document.elementFromPoint(%1, %2);
                 var morphContextMenuIsDocumentElement = false;
                 if (morphElemContextMenu === null)
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -252,7 +252,6 @@ WebView {
     function returnIfPossible(certificateError){
         // deferred rejection so that c.v.Dialog isn't called twice
         certificateError.rejectCertificate
-
         // check if back is available
         if (currentWebview.canGoBack) 
         {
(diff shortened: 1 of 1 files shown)


UniversalSuperBox (changes requested):
This review should block the merge of this until we can figure out the "going back twice" problem.

I wonder if the problem is that you're connecting to `rejectCertificate`. What does the handler for that signal do?

njtwiseman:
So just to clarify that we're both on the same page (haha) 

I start on: 
ubports.com > badssl.com > click on the "Expired" button
When I reject the tab goes away a second or two passes and then the browser returns to 
ubports.com < 2 (We want to be here?) <  1

I think `rejectCertificate` sends something into the background of QML / the webView - haven't been able to find out yet.  Documentation says it _must_ be called when you revoke.

It also says, after reading through it a second time, that it Ends the defer() command. Which, at this point, has paused the URL request.  Upon receiving either the `ignoreError` or reject signals The url request is either continued for ignore or dropped for reject.  

Which makes me think the back command is unnecessary. Unless there's more than one request included in the bad certificate request response exchange -> leaving you in no man's land without a place to go and not having a place to be either.

njtwiseman:
Well that wasn't so bad

njtwiseman (commented):
Seems to be working at this time.

I suspect `certificateVerificationDialog.reject` being called twice was the problem.

src/app/WebViewImpl.qml
```diff
@@ -248,18 +248,32 @@ WebView {
...
+        certificateVerificationDialog.host = UrlUtils.extractHost(certificateError.url);
+        certificateErrorsMap[certificateVerificationDialog.host] = certificateError
+        certificateErrorsMapChanged()
+        certificateVerificationDialog.localizedErrorMessage = certificateError.description
+        certificateVerificationDialog.errorIsOverridable = certificateError.overridable
+        certificateVerificationDialog.accept.connect(certificateError.ignoreCertificateError)
+        certificateVerificationDialog.reject.connect(returnIfPossible)
+        
+        
+    }
```
UniversalSuperBox, line 272: This many newlines seems like overkill

UniversalSuperBox (changes requested):
Only a style change, it seems to function correctly.

src/app/WebViewImpl.qml
```diff
@@ -313,7 +306,7 @@ WebView {
                     quickMenu.visible = true;
                 }
 
-                var commandGetContextMenuInfo = "
+                var commandGetContextMenuInfo = ""
```
njtwiseman: This didn't look intended so I added a close quote

src/app/WebViewImpl.qml
njtwiseman: yup

njtwiseman:
I see the problem from #167, incorporated upstream changes -> should be good

src/app/WebViewImpl.qml
```diff
@@ -248,35 +248,43 @@ WebView {
...
-          certificateErrorsMapChanged()
-          certificateVerificationDialog.localizedErrorMessage = certificateError.description
-          certificateVerificationDialog.errorIsOverridable = certificateError.overridable
-          certificateVerificationDialog.accept.connect(certificateError.ignoreCertificateError)
-          certificateVerificationDialog.reject.connect(certificateError.rejectCertificate)
-      }
+    // returnIfPossible -> rejects bad certificate and returns to last page
+    function returnIfPossible(certificateError){
+        // deferred rejection so that c.v.Dialog isn't called twice
+        certificateError.rejectCertificate
```
balcy, line 254: this line shows `Type Error: Cannot read property 'rejectCertificate' of undefined` in the logfile.
Since QtCreator was showing a warning, I changed it to `certificateError.rejectCertificate()`
But then it was still not working: `TypeError: Cannot call method 'rejectCertificate' of undefined`
so in fact "certificateError" is not passed to the function "returnIfPossible", and does nothing

balcy: ... but before putting more effort into it, we should discuss with @UniversalSuperBox and @cibersheep  if maybe a Cancel button is enough, and we could instead add the back button to the error sheet later. (https://github.com/ubports/morph-browser/issues/116)

njtwiseman: So for some reason `certificateError` isn't getting passed to the function.

balcy: I think only arguments passed to the reject() signal of the dialog would get passed to the function

balcy:
instead of automatically navigating back from the certificate error dialog, the "back to safety" button is put to the ErrorSheet instead (see comments in the issue)

PR #177, opened 2019-04-19
use currentWebview for PopupUtils

balcy:
maldito found that that the recent change broke the export functionality.
(the additional parameter has been added to make it work on private windows. instead of 'this'  use currentWebview.


myii:
@balcy Thanks for this, I've applied this fix manually to my device and can confirm that both types of export are now working!

UniversalSuperBox (approved):

PR #185, opened 2019-05-08
integrate certificate error "backToSafety" button to the ErrorSheet

### LINKED ISSUE #116: "Back to Safety" on certificate error dialog does not go back

balcy:
fixes https://github.com/ubports/morph-browser/issues/116

### DIFF
--- a/src/Morph/Web/MorphWebView.qml
+++ b/src/Morph/Web/MorphWebView.qml
@@ -304,6 +304,7 @@ WebEngineView {
         id: internal
         property int lastLoadRequestStatus: -1
         property string lastLoadRequestErrorString: ""
+        property int lastLoadRequestErrorDomain: -1
         property QtObject contextModel: null
 
         function dismissCurrentContextualMenu() {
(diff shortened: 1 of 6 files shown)


UniversalSuperBox (commented):
Sorry for the delay...

After clicking "Cancel" on the certificate error dialog, the error page is shown offering to take the user back to safety or try refreshing the page.

image

This seems to be what was discussed in #116. If @cibersheep approves of this design, I'll merge it.


By going to badssl.com/dashboard, we can see that behavior for bad SSL requests coming from a page has not changed: The attempted navigation causes a user-visible error. This is not expected (the errors should probably be silently denied), but again, it is not a regression.

While trying to test this, I went to check out my favorite "broken SSL" site, allrecipes. Instead of throwing a ton of SSL errors like it normally did, the browser went into an endless loop trying to flash the error page. Not sure what's going on and the site loads over 500 resources between all of its JavaScript and the things which the JavaScript load in turn... Makes debugging almost impossible and the issue doesn't seem to be caused by these changes anyway.

UniversalSuperBox (approved):

PR #203, opened 2019-06-23
Xenial   domainpermissions

### LINKED ISSUE #147: Problems with custom URL schemes (e.g. for telegram)
Currently there is a problem with some pages that make requests to custom URL schemes without user interaction.
Example: https://t.me/ubports/176342

The page will end up flickering with the URL chrome-error://chromewebdata/

The default policy for unknown schemes is to allow them when initiated via user interaction, but not to allow them for other cases. (like the policy for popup windows)

https://doc.qt.io/qt-5/qml-qtwebengine-webenginesettings.html#unknownUrlSchemePolicy-prop

The easiest solution to that would be adding 
`settings.unknownUrlSchemePolicy: WebEngineSettings.AllowAllUnknownUrlSchemes`
to WebViewImpl.qml

But I'm not sure if we intend this for morph... (in this case opening the example URL would immediately open the telegram app, or switch to it if it is already open)
Another thing I've tried is to catch this specific error in the loadingChanged event:
in Morph/Web/MorphWebView.qml I've added the following at the beginning of onLoadingChanged:
```
 if (loadRequest.errorCode === 420) {
    _webview.stop()
    return
 }
```
with that the example page does open, and the telegram app would only open if you click on "VIEW IN GROUP".
But with that the example page does not always show the post:
image
Sometimes it will display like on the left, sometimes like on the right. (so the user has to refresh the page to see the post)
This problem does not appear, if we change the policy as written above

### LINKED ISSUE #154: save domain specific settings

### LINKED ISSUE #195: Chrome-error   chromewebdata
You can try this : 

Go to https://leboncoin.fr to see an advertisement.

Try to see the phone number of the vendor in pressing the blue  phone number zone.

screenshot20190610_000824814

You must see very fleetingly the phone number but just after the screen sparkle to see a white chromewebdata page and through a black network error page : problems to display : chrome-error://chromewebdata/.

screenshot20190610_000949480
screenshot20190610_000943219

### LINKED ISSUE #200: block / whitelist domains
while doing some debugging I showed a message box for all requests of a webapp.
Many requests came from 3rd party domains. With a desktop browser and an addon like noscript you can easily block the ones not needed for just using the site normally.
I thought it would be useful to have control about them in morph, too.
With the "DomainPermissions" it is possible to block domains, or even opt in to a "whitelist mode" that requires each domain to be confirmed once.

balcy:
This PR adds 2 new features:
1) domain settings
fixes https://github.com/ubports/morph-browser/issues/154
fixes https://github.com/ubports/morph-browser/issues/147
fixes https://github.com/ubports/morph-browser/issues/195

2) domain permissions
fixes https://github.com/ubports/morph-browser/issues/200

both are added for morph-browser and webapps


balcy:
hints for testing:

domain settings:
- navigate to https://t.me/ubports/176342
- in the dialog, tick the checkbox "save permanently" and press OK
- the telegram app opens
- in Settings -> Domain Specific Settings "t.me" with the tick "allow custom schemes" is present
- restart morph and visit the URL again, the telegram app opens automatically

zoom:
- navigate to some page, e.g. the initial page
- select "Zoom" from the main menu
- set the zoom to a value that is not the default zoom
- in Settings -> Domain Specific Settings the domain entry with the selected zoom appears
- restart morph and visit the page again.
- select "Zoom" from the main menu, the zoom is the one selected last time, and the page zoom matches that value

location:
- activate location access on the phone
- navigate to a page asking your location, e.g. https://www.w3schools.com/html/tryit.asp?filename=tryhtml5_geolocation
- in this example click on "Try it"
- in the Permission Request, check the "Save permanently" checkbox, and press "Allow"
- your current location is displayed in the box at the bottom of the page
- in Settings -> Domain Specific Settings the entry "www.w3schools.com" (or the page you used) exists and the checkbox "allow location access" is checked
- restart morph and visit the page again
- after pressing "try it" the page does not ask for the location permission this time, and your location is displayed at the bottom of the page

... custom user agents will follow soon

balcy:
~~to test the custom user agents:~~
* ~~navigate to https://zn6kx.codesandbox.io/ (user agent page 1) and save it to the favorites~~
* ~~navigate to https://xcdqd.codesandbox.io/ (user agent page 2) and save it to the favorites~~
* ~~for both pages the default (desktop or mobile) user agent should appear~~
* ~~go to Settings -> Domain Specific Settings -> Custom User Agents~~
* ~~click "+", type "u1" as name for the new user agent, and then click "OK"~~
* ~~in the textarea type "dummy user agent 1 ..."~~
* ~~click "+", type "u2" as name for the new user agent, and then click "OK"~~
* ~~in the textarea type "dummy user agent 2 ..."~~
* ~~go to Settings -> Domain Specific Settings~~
* ~~click "+" and type "zn6kx.codesandbox.io" to add the domain (alternatively, change the zoom on the page as described above to add it automatically)~~
* ~~check the box "custom user agent"~~
* ~~click on "u1"~~
* ~~click "+" and type "xcdqd.codesandbox.io" to add the domain (alternatively, change the zoom on the page as described above to add it automatically)~~
* ~~check the box "custom user agent"~~
* ~~click on "u2"~~
* ~~navigate to https://zn6kx.codesandbox.io/~~
* ~~you should see "dummy user agent 1 ..."~~
* ~~navigate to https://xcdqd.codesandbox.io/~~
* ~~you should see "dummy user agent 2 ..."~~

**note:** there seems to be one issue with the page history (back and forward): by changing the user agent it seems to be lost, maybe because it is saved in the profile and assumed to stay constant per webview by QtWebEngine, or it is a problem of how the user agent is changed in morph, not sure here...

balcy:
I've deactivated the custom user agents for now, because I've not found a way to reliably change them for specific domains.
An alternative to changing httpUserAgent of the WebEngineProfile might be https://doc.qt.io/qt-5/qquickwebengineprofile.html#setUrlRequestInterceptor, see https://github.com/qutebrowser/qutebrowser/issues/3636

UniversalSuperBox:
Because of the size of this PR, I think merging it will unfortunately need to be held until after the release of OTA-10. All the review and testing I can do might not reveal how the changes interact with each other, and we're a little close to release. There are also some string changes and the translators might hit me with their devices if I merge this so late. :P

UniversalSuperBox:
Wow, how close we were to release indeed :roll_eyes:

Review starts *now*!

src/Morph/Web/MorphWebView.qml
```diff
@@ -322,6 +322,9 @@ WebEngineView {
     readonly property bool lastLoadFailed: internal.lastLoadRequestStatus === WebEngineLoadRequest.LoadFailedStatus
     readonly property string lastLoadRequestErrorString: internal.lastLoadRequestErrorString
     onLoadingChanged: {
+        if (loadRequest.errorCode === 420) {
```
UniversalSuperBox, line 325: Is this something thrown internally by QtWebEngine or do you use a lot of Spring?

src/app/AllowCustomSchemesDialog.qml
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2019 ubports
```
UniversalSuperBox: We are currently unable to have copyright assigned to us. You can keep your copyright if you want.

src/app/AllowCustomSchemesDialog.qml
```diff
@@ -0,0 +1,61 @@
...
+    objectName: "confirmDialog"
+    title: i18n.tr("Custom URL schemes")
+
+    property string url
+    property string domain
+    property bool showAllowPermanentlyCheckBox
+
+    message: i18n.tr("The following url with a custom scheme was blocked:") + "\n" +
+                                           url + "\n\n" +
+                                           i18n.tr("Should all custom URL schemes from domain %1 be allowed?".arg(domain));
```
UniversalSuperBox: This spacing seems random. Why so far to the right?

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                }
+            },
+            Action {
+                iconName: "add"
+                visible: !selectMode
+
+                onTriggered: {
+                    var promptDialog = PopupUtils.open(Qt.resolvedUrl("PromptDialog.qml"), domainPermissionsPage);
+                    promptDialog.title = i18n.tr("Add domain")
+                    promptDialog.message = i18n.tr("Enter the name of the domain, e.g. example.com (subdomains will be removed).")
```
UniversalSuperBox: May this dialog also have a "Current page" option, to add domain settings for the site you're currently viewing?

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                    promptDialog.accept.connect(function(text) {
+                        if (text !== "") {
+                            var domain = DomainPermissionsModel.getDomainWithoutSubdomain(UrlUtils.extractHost(text));
+                            if (DomainPermissionsModel.contains(domain)) {
+                                domainPermissionsItem.setDomainAsCurrentItem(domain);
+                            }
+                            else {
+                                DomainPermissionsModel.insertEntry(domain, false);
+                                reload(domain);
+                            }
```
UniversalSuperBox, line 126: Wait, what are we doing if the user enters nothing? Does `domain` exist here?

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                onTriggered: {
+                    selectMode = true
+                }
+            },
+            Action {
+                iconName: "add"
+                visible: !selectMode
+
+                onTriggered: {
+                    var promptDialog = PopupUtils.open(Qt.resolvedUrl("PromptDialog.qml"), domainPermissionsPage);
```
UniversalSuperBox, line 113: Does `PromptDialog` allow giving the text box a URL type? It seems strange to have it listed as a normal text box. Also makes autocorrect go nuts.

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+            }
+
+            ViewItems.selectMode: selectMode
+
+            delegate: ListItem {
+                id: item
+                readonly property bool isCurrentItem: item.ListView.isCurrentItem
+                readonly property string domain: model.domain
+                height: isCurrentItem ? layout.height : units.gu(5)
+                color: isCurrentItem ? theme.palette.selected.base : theme.palette.normal.background
```
UniversalSuperBox: `base` is quite dark now. I suspect it is not the color you want.

src/app/DomainSettingsPage.qml
```diff
@@ -0,0 +1,390 @@
...
+                            font.bold: item.ListView.isCurrentItem
+                        }
+
+                        Row {
+                            spacing: units.gu(1.5)
+                            height: units.gu(1)
+                            visible: item.ListView.isCurrentItem
+
+                            Label  {
+                                text: i18n.tr("allow custom schemes")
```
UniversalSuperBox: "Allow this site to launch other apps on your device"

src/app/DomainSettingsPage.qml
```diff
@@ -0,0 +1,390 @@
...
+                        }
+
+
+                        Row {
+                            spacing: units.gu(1.5)
+                            height: units.gu(1)
+                            visible: item.ListView.isCurrentItem
+
+                            Label  {
+                                text: i18n.tr("allow location access")
```
UniversalSuperBox: "Allow this site to access your location"

src/app/DomainSettingsPage.qml
```diff
@@ -0,0 +1,390 @@
...
+                            }
```
UniversalSuperBox, line 224: Is it appropriate to have location settings here? I think that "Settings -> Privacy and Permissions" already contains camera and microphone permissions, so it should also have these.

src/app/webbrowser/SettingsPage.qml
```diff
@@ -268,6 +284,36 @@ FocusScope {
                         id: privacyCol
                         width: parent.width
 
+                        ListItem {
+                            objectName: "setDomainWhiteListMode"
+
+                            ListItemLayout {
+                                title.text: i18n.tr("Domain whitelist mode")
```
UniversalSuperBox: "Only allow browsing to whitelisted websites"

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        Label {
+            id: emptyLabel
+            anchors.centerIn: parent
+            visible: domainPermissionsListView.count == 0
+            wrapMode: Text.Wrap
+            width: parent.width
+            horizontalAlignment: Text.AlignHCenter
+            text: i18n.tr("No domain specific permissions available")
```
UniversalSuperBox: "No sites have been granted special permissions"

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                                visible: (model.requestedByDomain !== "")
+                                text:  "(→%1)".arg(model.requestedByDomain)
+                                color: theme.palette.normal.foregroundText
+                            }
+
+                        }
+
+                        ColumnLayout {
+                            visible: item.ListView.isCurrentItem
+                            CustomizedRadioButton {
```
UniversalSuperBox, line 221: I'm unable to select any of these radio buttons upon creating a site. I created `example.com` while on a different site, then tried to select a radio button. No dice.

src/app/DomainSettingsPage.qml
```diff
@@ -0,0 +1,390 @@
...
+
+            ViewItems.selectMode: selectMode
+
+            delegate: ListItem {
+                id: item
+                readonly property bool isCurrentItem: item.ListView.isCurrentItem
+                readonly property string domain: model.domain
+                readonly property int userAgentId: model.userAgentId
+                height: isCurrentItem ? layout.height : units.gu(5)
+                color: isCurrentItem ? theme.palette.selected.base : theme.palette.normal.background
```
UniversalSuperBox: Again, I think that `base` is much darker than you expect.

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                                onCheckedChanged: {
+                                    if (checked) {
+                                        DomainPermissionsModel.setPermission(model.domain, DomainPermissionsModel.Blocked, false)
+                                    }
+                                }
+                            }
+
+                            CustomizedRadioButton {
+                                checked: (model.permission === DomainPermissionsModel.Whitelisted)
+                                text: i18n.tr("Whitelisted")
```
UniversalSuperBox: "Always allow access"

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+                                color: theme.palette.normal.foregroundText
+                                onCheckedChanged: {
+                                    if (checked) {
+                                        DomainPermissionsModel.setPermission(model.domain, DomainPermissionsModel.NotSet, false)
+                                    }
+                                }
+                            }
+                            CustomizedRadioButton {
+                                checked: (model.permission === DomainPermissionsModel.Blocked)
+                                text: i18n.tr("Blocked")
```
UniversalSuperBox: "Never allow access"

src/app/DomainPermissionsPage.qml
```diff
@@ -0,0 +1,290 @@
...
+    signal done()
+    signal reload(string selectedDomain)
+
+    BrowserPage {
+        id: domainPermissionsPage
+
+        anchors.fill: parent
+        focus: true
+
+        title: i18n.tr("Domain specific permissions")
```
UniversalSuperBox: It's unclear what a domain specific permission is. Is "Website Blocking" more accurate?

UniversalSuperBox (changes requested):
Here's what I've found so far. I believe we have some good bones for functionality here, but it's unclear what some of the settings do.

Paging @cibersheep to try this out too (`sudo ubports-qa install xenial_-_domainpermissions`). There's a lot of new settings, menus, and buttons to play with.

src/app/DomainPermissionsPage.qml
UniversalSuperBox: Or is "Domain blacklist/whitelist" better

cibersheep:
> Paging @cibersheep to try this out too (`sudo ubports-qa install xenial_-_domainpermissions`). There's a lot of new settings, menus, and buttons to play with.

Luvely. I'll have something to do at the start of the week :)

cibersheep:
Installed. Were I should look for the changes?

UniversalSuperBox:
Tap through Settings a bit. There are *lots* of new options.

I think there are some dialogs to be hit when browsing sites with bad behavior, but I haven't seen them yet.

giiba:
I did the install as noted above and like Cibersheep I see no new options... are we sure that instal method works? I had to temporarily mount the apt archive in userdata to make it work so I am confident something was installed but none of the changes seem to be in effect.

cibersheep:
Weird. After installing, I see no changes in the Settings menu

UniversalSuperBox:
I've triggered a new build in Jenkins. After it finishes, run `sud ubports-qa update` to get the new goodies.

giiba:
That did the trick. This is wonderful and solves a few of the major gripes I had with Morph!

At first pass things seem to be working well, nice job.

cibersheep:
Nice. Domain Permissions settings needs a little tweak but I think is good for now.

aka: 
- more space between option «allow...» with check boxes
- checkboxes to be aligned to the right

src/Morph/Web/MorphWebView.qml
balcy: yes it is internally thrown by QtWebEngine 5.11:
https://github.com/qt/qtwebengine/blob/5.11/src/core/web_contents_delegate_qt.cpp
lines 598 and 600.
But they are no longer using it for 5.13
Ah now I get it: spring framework, no never used that : )
If the custom schemes are not yet allowed, you see a flickering (always reloading) page of the second screenshot of https://github.com/ubports/morph-browser/issues/195
So we can probably remove this again with QtWebEngine 5.13.

src/app/DomainPermissionsPage.qml
balcy: there is the check `if (text !== "") {` in line 119

src/app/DomainPermissionsPage.qml
balcy: sounds good, I think the settings page cannot directly access this information though. (we might need to add another property to the settings page)
for the zoom, location and custom url schemes there is already a workflow that adds the current page as needed. (e.g. if user changes the zoom or the page needs location access)

src/app/DomainPermissionsPage.qml
balcy: yeah that is a good point, I try adding that property to the prompt dialog.

src/app/DomainPermissionsPage.qml
balcy: yes I think that was a trade-off where I tried to make it look acceptable with light and dark theme, corrections welcome : )

src/app/DomainSettingsPage.qml
balcy: yeah might need tweaking

src/app/AllowCustomSchemesDialog.qml
balcy: could we use "ubports contributors" instead ?

src/app/AllowCustomSchemesDialog.qml
balcy: no reason, I'll remove the additional spaces.

src/app/AllowCustomSchemesDialog.qml
UniversalSuperBox: No, just yourself

src/app/DomainPermissionsPage.qml
UniversalSuperBox: Ah, okay.

src/app/DomainPermissionsPage.qml
balcy: uh, that is strange, can it be because I am on the edge channel ? Is the "Not Set" radio button selected after you create "example.com" ? For me if I click on the radio buttons or the texts "Blocked" or "Whitelisted", the selection changes.

src/app/DomainPermissionsPage.qml
balcy: agreed, maybe the 2nd one.

src/app/DomainSettingsPage.qml
balcy: I think it was done to have all domain specific settings in one place. (e.g. adding, editing, and removing them), as they are in a common data model.
would it help to put the whole category "domain settings" to "privacy and permissions" ?

balcy: b.t.w. the microphone and camera permissions are not functional at the moment

UniversalSuperBox: Moving them would make sense. A little unfortunate about mic and camera.

src/app/DomainPermissionsPage.qml
cibersheep: It happens only if: 
- there is no previous information stored (like used for the first time) 
- go to settings / domain permisions
- add a url
- tap on select

src/app/AllowCustomSchemesDialog.qml
balcy: there are examples of "the UBports project" though:
https://github.com/search?p=4&q=org%3Aubports+%22The+UBports+project%22&type=Code

src/app/DomainPermissionsPage.qml
balcy: I removed the database (/home/phablet/.local/share/morph-browser/domainpermissions.sqlite) to have the situation of the first use.
I see the following after adding the url:

image

what is "tap on select" referring to ?

cibersheep: Sorry. You might want to read «Tap on select» as «Tap on edit» :D 
Mmm... The menu I see is different, I might not have your latest changes :)

src/app/AllowCustomSchemesDialog.qml
balcy: is it OK if I remove all lines "Copyright 2019 ubports" altogether, and put no copyright to the single files at all ?

src/app/DomainSettingsPage.qml
balcy: was a bit long, have used "allowed to launch other apps"

src/app/DomainSettingsPage.qml
balcy: was a bit long, have changed it to "allowed to access your location"

src/app/AllowCustomSchemesDialog.qml
UniversalSuperBox: Maybe if the whole project was that way, but in this case there's a convention to follow.

balcy: > We are currently unable to have copyright assigned to us. You can keep your copyright if you want.

ok but the convention was "Copyright 2013-2017 Canonical Ltd." and I don't want to keep the copyright

src/app/DomainPermissionsPage.qml
balcy: @UniversalSuperBox @cibersheep  unfortunately I cannot reproduce that, could you maybe add a screenshot,or does the log say sth?
If you activate the checkbox "domain whitelist mode" (old version) / "only allow browseing to whitelisted websites" (new version), does morph properly ask for each new domain if it should be blocked or whitelisted ? Are these settings reflected in the domain permissions menu ?
[edit] with Aquaris 4.5 and Desktop the radio boxes do work for me

src/Morph/Web/MorphWebView.qml
UniversalSuperBox: Sounds good

src/app/AllowCustomSchemesDialog.qml
balcy: I could just write my name for now, and later if the foundation is ready, change it back maybe ?

src/app/DomainPermissionsPage.qml
balcy: this feature I would rather skip for now, as soon as custom user agents will (hopefully) be activated again, it will become handy to have it.
For now the location / custom schemes / zoom do already have a workflow to add the current page automatically.

src/app/DomainPermissionsPage.qml
UniversalSuperBox: Hmm. I originally reproduced that on my Oneplus One across a few restarts of the browser and phone, but I can't do it any longer. Guess we'll keep watch for that one.

src/app/DomainPermissionsPage.qml
UniversalSuperBox: Ah, don't worry about it for now then.

src/app/AllowCustomSchemesDialog.qml
```diff
@@ -0,0 +1,61 @@
+/*
```
UniversalSuperBox, line 1: Hmm. Functionally, this dialog is fine. But I wonder if we can move it around and change the text to make it clearer. @cibersheep, do you have any thoughts?

screenshot20190915_174200521

src/app/DomainPermissionsPage.qml
cibersheep: It looks I cannot reproduce it anymore. I guess after rebooting a couple of times is gone. Mmmm....

src/app/AllowCustomSchemesDialog.qml
cibersheep: Yep. I think is a bit confusing because of the structure.
I got there doing:
 - going to t.me[blbalba]
- Allow/Block dialog t.me. Tapped allow
- Custom URL schemes dialog saying «t.me is blocked»

In this Dialog, what about something like this?:

**Custom URL Scheme**
This site is trying to ¿open? the URL scheme: `tg:[blablab]`

This will try to dispatch this url to a system app if possible. Do you want to proceed?

Always allow URL schemes from t.me []

[allow]
[Cancel]

cibersheep: So, first I get 

imatge

then 

imatge

cibersheep:
About the design tweaks. I have some thoughts:

On the Allow/Block Dialog

imatge

I would choose a main action and color only that one.

**Block Domain**
About to access to domain XXX. Do you want to block this domain from now on?

[Block domain] (in red)
[Allow] (uncolored)
[Cancel] (uncolored)

Or the opposite, if you think the main action is to allow domains_

[Allow] (in green)
[Block domain] (uncolored)
[Cancel] (uncolored)

On Dialogs like this:
imatge

I would show a specific text in the button. More like:
**This is happening**
Do you want to [action]?
[Action]
[Cancel]

And reserve the "Ok" answer for informative texts:
**This happened**
This is what happened right now.
[Ok]

In the Domain Specific Settings:

imatge

- I would go for a clear grey (Silk, even Porcelain)
- Center the check boxes to the right with ore space (ListItems even)

In the Domain Specific Permissions:

imatge

- Same with grey
- I would show Not set, Blocked and Whitelisted in `theme.palette.normal.backgroundText`
- I would use an exposed [combo button](https://api-docs.ubports.com/sdk/apps/qml/Ubuntu.Components/ComboButton.html) to show the options
- And use the Suru icons for the allowed and blocked icons :)

If you want me to help with any / all those, ping me

src/app/AllowCustomSchemesDialog.qml
UniversalSuperBox: For reference, here are some examples of the same dialog from Chromium and Firefox:

image

Screenshot at 2019-09-16 10-30-23

Of course we can't select the application to launch, nor would the browser know which application is going to be launched.

balcy:
- the background color has been adapted (commit 3 days ago)
  (the color is derived from the background, but relatively lighter or darker, depending on the theme, feel free to change / suggest the factor here)
- the alignment of the checkboxes has been made (commit 3 days ago)
- the custom url scheme dialog has been adapted (current commit)
  [the text for the checkbox does not include the domain, because it could be too long]
- the blacklist/whitelist page has been adapted, so that the radio box labels are no longer bold and no longer green / red

for the "allow block" dialog I think neither allow, nor block is the default action:
Depending on the page and what the user wants, they have to take the decision.
And both selections result in a blacklist or whitelist entry in the database. (the same page might include content from desired and non-desired domains)

- I think radio boxes for an enum with only 3 fixed values are easier to use than combo boxes
- with the suru icons I'm not sure which one could be better, I've used color overlays (green / red) to make the list easier to read.

src/app/AllowCustomSchemesDialog.qml
cibersheep: Firefox message seems more understandable (+1 FF)

cibersheep:
Coool... let me try your changes and I get back to you.

Meanwhile:
- Suru Icons you might use the ones named: "ok" for allowed, "cancel" for blocked. And they can be colored as you wish
- About the Allow/Deny Dialog... I agree that is tricky. What about following the Restart menu style?
Meaning: only one of the options is colored (red or green accordingly) the user still needs to choose (and not forcing block or allow as I suggested)

imatge

- Radio buttons. The thing is I can't remember any where in the system we use them. We use ListItems for that:

ie.
GSM provider:
imatge

Lock screen after x seconds:
imatge

balcy:
- icons: ah they really look much better, especially in the dark mode (would you keep dialog-question-symbolic for "NotSet" ?

- Allow / Deny dialog: Let's say the default would be allow (green), because most pages the user navigates to would be allowed. But if "block" is grey then, the user might not know that he will no longer be able to access the page from then.
If we really have to avoid having a "positive" and "negative" button in one dialog, I'd prefer "Allow" grey, and "Block" red. But I like the current version most, because the colors reflect the status in the permissions page.

- the radio boxes: the examples from the settings page are selected items from a list, but here we select an option of a single list item.
While we already highlight the list item (the domain), the "sublist" item would need another color again to see the selcted item. Also, the selected item would have a tick on the right, that looks like the symbol to whitelisted items. I would still vote for those here, if many users complain we can change them : )

cibersheep:
Thank you for your work, I have tested the updates you did :)

- The info icon for Not set could work until I can design something specific (if you like)
- For the Allow / Deny Dialog, I know that is a bit special but I really think we should be quite meticulous with the core apps therefore I would go with the top as a red block button
- And for the radio buttons: I see what you mean. What about leaving it like it is for now and come back to it in a later revision?

balcy:
>     * The info icon for Not set could work until I can design something specific (if you like)
ok thanks, for now I'm fine with the icon
> 
>     * For the Allow / Deny Dialog, I know that is a bit special but I really think we should be quite meticulous with the core apps therefore I would go with the top as a red block button
ok I tried to change that
>     * And for the radio buttons: I see what you mean. What about leaving it like it is for now and come back to it in a later revision?
it is not carved in stone : )

balcy:
@UniversalSuperBox  ok ready for merge ?

src/app/GeolocationPermissionRequest.qml
```diff
@@ -23,29 +23,37 @@ import Ubuntu.Components.Popups 1.3
...
+    onAllowPermanently: { PopupUtils.close(dialog); }
+    onReject: { PopupUtils.close(dialog); }
 
-    Button {
-        objectName: "deny"
-        text: i18n.tr("Deny")
-        onClicked: reject()
+    ListItemLayout {
+        visible: showAllowPermanentlyCheckBox
+        title.text: i18n.tr("Save permanently")
```
UniversalSuperBox: Since this was changed in the custom URL scheme request, it should also be changed to "Remember decision" here.

src/app/AllowCustomSchemesDialog.qml
```diff
@@ -0,0 +1,61 @@
...
+
+    ListItemLayout {
+        visible: showAllowPermanentlyCheckBox
+        title.text: i18n.tr("Remember decision")
+        CheckBox {
+            id: allowPermanentlyCheckBox
+         }
+    }
+    Button {
+        text: i18n.tr("Allow")
```
UniversalSuperBox, line 50: I noticed that allowing or denying one of these is remembered until the end of the session. That is, if I deny a URL open without checking "Remember decision" then click the URL again, I'm not asked to perform the action again.

Now, this is good for not spamming the user. Is it what you expected to happen?

src/app/AllowOrBlockDomainDialog.qml
```diff
@@ -0,0 +1,64 @@
...
+import Ubuntu.Components 1.3
+import webbrowsercommon.private 0.1
+
+ModalDialog {
+    objectName: "allowOrBlockDomain"
+    title: i18n.tr("Block domain")
+
+    property string domain
+
+    message: i18n.tr("About to access domain %1. Do you want to block this domain from now on?".arg(domain));
```
UniversalSuperBox: This text still seems unclear to me. As a user, I don't know what will happen if I choose each option. Since this dialog only appears when whitelist-only mode is enabled (I think), let's refine it as "You're trying to access *domain* but it is not on your approved domain whitelist. Would you like to continue to block the domain, allow browsing to the domain, or go back?"

UniversalSuperBox (changes requested):

src/app/AllowCustomSchemesDialog.qml
balcy: yes this is done on purpose. Since the for geolocation requests this behavior (remembering the decision for the current session) is already done by the backend (QtWebEngine), I wanted the same behavior here.

By the way: in incognito mode the "remember decision" checkboxes are not available.

src/app/AllowOrBlockDomainDialog.qml
UniversalSuperBox: Maybe that's too long.

src/app/AllowCustomSchemesDialog.qml
UniversalSuperBox: Okay, just making sure.

src/app/AllowOrBlockDomainDialog.qml
balcy: yes it is right, the dialog does only appear in "whitelist-only" mode.

There are two types of requests though: mainFrame requests (navigating to a different page) and requests within a page to another domain.

For the non-mainFrame requests I would suggest sth. like: 
"_parentDomain_ is trying to access _domain_ but it is not on your approved domain whitelist. Would you like to continue to block the domain, allow loading data from the domain, or just ignore this request ?"

UniversalSuperBox: How would I see the second type of request?

balcy: e.g https://nasa.gov
those requests are async, so several dialogs will appear. But the order the user blocks them does not matter.

cibersheep: Mmm... tested the latest changes. I think the experience could be polished a bit but I don't think is bad how it is now. How about going ahead how it is and revisit later on?

balcy: by the way: the allow / block dialog was inspired from the "Jade Diamond" browser : )
image

cibersheep: Jade Diamond is not a core app  ;)

UniversalSuperBox (approved):
Here we go.

giiba:
Nice work folks!

I've been testing the webapp and site specific zoom settings and it is really nice to have as it solves some of the problems that required more complicted solutions for zoom settings.

One curiosity: is there a way to apply zoom settings earlier in the rendering process? At the moment the pages are rendered then displayed, and then everything shifts afterwards as the content is fit to the zoom setting. This makes for a slightly awkward user experience as after a page loads things start moving before settling down.

giiba:
One thing I discovered this morning:

If a website asks for permission to location information the dialog won't let me disallow and remember the choice. If I check the 'remember decision' option the deny button is disabled.

Is there a reason for this? There are sites that ask it every page load, and I would like to just block the request permanently, as opposed to having to choose every time.

screenshot20190930_111828902

balcy:
yeah good point. At that time it was implemented like a permission that can be revoked later (just a boolean flag "allow location access" in the database).
So permanent blocking was not thought of there, and would require a new field in the database.
[edit: actually "allow permanently" was renamed to "remember decision" later]

giiba:
Well it doesn't hinder the functionality, as Morph(stable) asks every time anyways on a site like that. Really the problem is the website.

balcy:
> One curiosity: is there a way to apply zoom settings earlier in the rendering process? At the moment the pages are rendered then displayed, and then everything shifts afterwards as the content is fit to the zoom setting. This makes for a slightly awkward user experience as after a page loads things start moving before settling down.

it might be possible to use WebEngineLoadRequest.LoadStartedStatus instead of WebEngineLoadRequest.LoadSucceededStatus, but we have to make sure that we still detect whenever the current domain changes. (WebViewImpl.qml, 879)

giiba:
I played a bit with WebEngineView.LoadStartedStatus but it doesn't seem to trigger. If the current LoadSucceededStatus is replaced the zoom is never applied, and if I add a fourth 'if' under onLoadingChanged for LoadStartedStatus (and copy+paste from LoadSucceededStatus) there is no difference in behaviour.

Might be futile, I don't know enough, and it really isn't a problem.

Issue #202, opened 2019-06-21
Unexpected behavior with full-screen browser


NeoTheThird:
The browser is showing some unexpected behavior on edge, it seems to be running in full-screen. This leads to the following issues:

- The indicator bar is not shown
- The browser can not be moved in a window in desktop mode
- The launcher and app drawer can not be opened in desktop mode
- The top bar shows up while dragging out the launcher and drawer and disappears again when the the drawer is fully extended, leading to weird shifting around
- Working with multiple windows is a pain in the ass
- Side stage sometimes works and sometimes doesn't
- Top bar disappears if browser is focused in side-stage and re-appears if focus is switched to the app in main stage

I don't see anything interesting going on in the logs for browser or unity8. Might be caused by ubports/unity8#144. Have not observed this behavior with other full-screen apps such as Pathwind.

UniversalSuperBox:
If you open a video, make it full screen, then take it out, does the behavior change?

NeoTheThird:
Yes it does, keen observation. Actually, if you watch a video on youtube for example and maximize it, it fills the full screen. But then as soon as the next video starts trough autoplay, the video only fills the website-area of the browser and does not cover the navigation and url bar. Closing the tab with the video does not change that behavior, neither does closing the browser app. The behavior i described above stays the same from the moment you open a video in full-screen until you either explicitly minimize a video or reboot the device.

PR #212, opened 2019-07-17
before saving the session before the last window is closed, call setFullscreen(false)

balcy:
fixes https://github.com/ubports/morph-browser/issues/202

### DIFF
--- a/src/app/webbrowser/morph-browser.qml
+++ b/src/app/webbrowser/morph-browser.qml
@@ -114,6 +114,7 @@ QtObject {
 
             onClosing: {
                 if (allWindows.length == 1) {
+                    allWindows[0].setFullscreen(false)
                     if (tabsModel.count > 0) {
                         session.save()
                     } else {


UniversalSuperBox:
Hmm, I don't think this problem is exclusive to the browser. Really I think it's a bug deeper in the stack, and individual apps should not have to work around it just because they happen to use fullscreen.

I'll move the bug report to unity8 so it isn't your "problem". Sorry I didn't do that before.

balcy:
ah ok, I thought the window state (that includes the fullscreen info) is saved and then later restored with that property.

UniversalSuperBox:
I believe it is, but I also believe it *shouldn't*

mariogrip:
Here is a global fix: https://github.com/ubports/unity8/pull/172

UniversalSuperBox:
Superseded by https://github.com/ubports/unity8/pull/172

PR #228, opened 2019-09-27
Xenial   downloadswebapps

### LINKED ISSUE #209: Download not working in webapp clicks
**Description of the feature**

    Device: N5, Bq 4.5, Fairphone, I think all
    Channel: Stable, RC, Devel
    Build: OTA 7, 8, 9

**Expected behavior**

- when clicking to download file at web-app, will open contenhub to select where to download and after that the file will be downloaded

**Actual behavior**

- open any web app, click on download and nothing happens, no changes, not downloading, loading not opening any content hub window

- opening the same web page directly with Morph without a web app it works well

**Logfiles and additional information**

- no logs to report
- do not know if it is only an issue due to created web app by clickable app or also using clickable desktop tool

balcy:
... enable downloads for webapps

fixes https://github.com/ubports/morph-browser/issues/209


src/app/WebViewImpl.qml
```diff
@@ -61,8 +61,6 @@ WebView {
     /*experimental.certificateVerificationDialog: CertificateVerificationDialog {}
     experimental.proxyAuthenticationDialog: ProxyAuthenticationDialog {}*/
 
-    signal showDownloadDialog(string downloadId, var contentType, var downloader, string filename, string mimeType)
```
giiba, line 64: Can we remove the restriction on certain contextual menu items so the context menu can trigger downloads too? Removing the '!isWebApp' should enable them I think.

See lines: 402, 417, 444, 460

Also consider lines 422 and 711.

I'd do the edit myself but I don't seem to be able to edit the files in a pull request (I'm new to this).

giiba:
So I have installed this PR and  tried my suggested changes. I can now choose save image, but the download window just comes up and says "no downloads availabe".

Clearly I know nothing...

However, I cannot actually trigger a download in any of the webapps I use (gmail being foremost of them). I plan to test others when I get a chance.

balcy:
seems to be a bug, I try to find it later

balcy:
@giiba the file "qmldir" (in /usr/share/morph-browser) was missing in the package, it should now show the (new) downloads. With `sudo ubports-qa update` you will get the corrected version

giiba:
I found that after installing the pr last night that the download function of Morph was broken; it would download files still, but no new entries would appear in the download list. It could be due to the missing files, I didn't investigate.

edit:
figured out the problem is on my end, I switched to RC to troubleshoot something else and it seems the install only works with the dev stream. Now I get downloads in my gmail webapp, so it appears to be working.

balcy:
since it is only one missing file, you could manually create / copy the file https://raw.githubusercontent.com/ubports/morph-browser/xenial_-_downloadswebapps/src/app/qmldir to "usr/share/morph-browser", too.
file: `/usr/share/morph-browser/qmldir`
content: `singleton ActiveDownloadsSingleton ActiveDownloadsSingleton.qml`

giiba:
Here is the output, I don't know how to solve this, or that it is the fault of this repo.

screenshot20190929_131905964

giiba:
Ignore my troubles above. With the update things seem to be working!

Is there a reason files are downloaded to ~/.local/share/[app x]/Downloads as opposed to just ~/Downloads?

giiba:
Aaaaaand, I tested my suggestion of enabling webapps to use context items like 'save link' and that is working. So I suggest those few '!isWebApp' conditions be removed.

balcy:
As I tested it the webapp-container did not have write permission for the Downloads folder.

giiba (dismissed):

giiba:
Makes sense. I've seen comments that apps should have access to the home directory, but my understanding was this is a broader operating system issue.

I foresee people downloading things and then asking in forums+chats "where did my files go?" or "downloads aren't working!".

Is there a way to bring up the downloads pane from within a webapp without starting a new download?

src/app/WebViewImpl.qml
balcy: 417, 444, 460 was adapted in the last commit

402 Actions.OpenLinkInNewPrivateWindow, that is not available for webapps (incognito mode)
422 / 711 : (share) maybe we can make a flag --enable-share for the webapp, because many webapps won't have the right permissions for that

balcy:
> I foresee people downloading things and then asking in forums+chats "where did my files go?" or "downloads aren't working!".
> 
> Is there a way to bring up the downloads pane from within a webapp without starting a new download?

yeah the only hint would be the folder written in the subtitle of the downloads page.
since the webapps don't have a menu, I've put an item to show the downloads to the webapps settings page now. (see latest commit)

src/app/WebViewImpl.qml
balcy: enabling share would require more changes, if I just remove the "!isWebApp", there is the error `file:///usr/share/morph-browser/WebViewImpl.qml:424: TypeError: Property 'shareLinkRequested' of object WebApp_QMLTYPE_77(0x12cd820, "webappBrowserView") is not a function` in the log

giiba:
That seems like a fine solution until such time they can just be placed in ~/Downloads.

UniversalSuperBox:
Do you have an example of this in action? Which webapps should I try and how should I access those settings?

balcy:
one example I've found is https://open-store.io/app/matecommunity.josele13
you can for example long press on the Ubuntu symbol top left, and then selct "save image" from the context menu.
The PNG then opens in the Downloads page, and you can open it in the gallery.

Deleting downloads from the list is possible like in morph-browser.

Without starting a new download, you can click on the settings icon top right, and then "Downloads".

For webapps without a top bar, those "webapp settings" (introduced with the domainsettings / domainpermissions), are accessible via the context menu (e.g. long press on the background of the webpage), and then the settings icon.

some of the articles have downloads, too.

maybe @MK1973  or others have example webapps ?

Any webapp should work, the only difference is that without the `content_exchange` permission, the content hub will show no apps to open the downloaded files.

giiba:
I use a custom simple gmail app on my phone, I have a seperate app for each of my accouts which nicely compartmentalizes things. Adding the content hub permission and this PR enables downloads.

I've attached a copy for testing if anyone cares.

src/app/webcontainer/webapp-container.qml
```diff
@@ -192,7 +192,12 @@ BrowserWindow {
             DomainPermissionsModel.whiteListMode = settings.domainWhiteListMode;
             DomainSettingsModel.databasePath = webappDataLocation + '/domainsettings.sqlite';
             DomainSettingsModel.defaultZoomFactor = settings.zoomFactor;
+            DownloadsModel.databasePath = webappDataLocation + "/downloads.sqlite";
             UserAgentsModel.databasePath = DomainSettingsModel.databasePath;
+
+            // this can be set from QtWebEngine version 1.9 (Qt 5.13)
+            //item.currentWebview.profile.downloadPath = webappDataLocation + "/Downloads";
+            FileOperations.mkpath(webappDataLocation + "/Downloads");
```
UniversalSuperBox, line 202: Could you also add the note that the part in WebApp.qml can be deleted when this is enabled? Maybe file an issue to fix both and reference that issue here.

UniversalSuperBox (changes requested):

UniversalSuperBox (approved):
Great as always. I especially appreciate your work to refactor these things as you go, removing dead or commented out code.

Issue #230, opened 2019-10-01
Readme outdated


lduboeuf:
Readme is not accurate with morph browser anymore.
Didn't had success to try morph with clickable. Any hints ?

lduboeuf:
ok just by putting CLICK_MODE to `on` in CMakeList went fine.
Not sure if the right path but i can propose a PR to add clickable instructions

PR #232, opened 2019-10-02
add instruction for building with clickable

lduboeuf:
refs #230 
convert readme to markdown
add clickable build args


balcy (dismissed):
would be good to add https://github.com/lduboeuf/morph-browser/pull/1 before merge

balcy (approved):

Issue #163, opened 2019-02-27
[feature request] implement delete cookies option


Danfro:
As far as I know cookies are stored in the sqllite database at: `.local/share/morph-browser/cookies`.
There is no option in morph browser to view and/or remove cookies

I really would like to have an option to remove them from within morph browser.
I would think settings/privacy would be a good place where to put that option.
Ideally one could view existing cookies and delete them all/individual. To start with an "clear all cookies" option would be good.

UniversalSuperBox:
Fixed by https://github.com/ubports/morph-browser/pull/260

Danfro:
E5 RC. Confirmed, morph now does have a "clear all cookies" option.

jonnius:
Confirmed on FP2, RC.

cibersheep:
N5 RC, works:
- browse to a web ask to store cookies
- accept them
- delete them with Morpho
- go back to the website

It will ask to accept cookies again

Bolly:
BQFHD RC, works:

- browse to a web ask to store cookies
- accept them
- delete them with Morph
- go back to the website

It will ask to accept cookies again

PR #234, opened 2019-10-12
add "Delete all cookies" to privacy settings

balcy:
implements basic feature request of https://github.com/ubports/morph-browser/issues/163


UniversalSuperBox (changes requested):
I think that [QWebEngineCookieStore::deleteAllCookies()](https://doc.qt.io/archives/qt-5.11/qwebenginecookiestore.html#deleteAllCookies) would be a much better option to use here. [QQuickWebEngineProfile::cookieStore()](https://doc.qt.io/archives/qt-5.11/qquickwebengineprofile.html#cookieStore) or [QWebEngineProfile::cookieStore()](https://doc.qt.io/archives/qt-5.11/qwebengineprofile.html#cookieStore) can be used to access the cookie store.

balcy:
hmm strange it worked with clickable... have to look into it

balcy:
or it built successfully now. The reason for the rebase was, that I wanted to bring in the new "SharedIncognitoContext" from https://github.com/ubports/morph-browser/pull/234/commits/07572e028e575892ff40a7d285b7214fbfd3500e.
the cookies from the standard profile, and the incognito profile are both deleted.

src/app/webcontainer/WebappSettingsPage.qml
```diff
@@ -202,6 +203,17 @@ FocusScope {
                                 dialog.confirmed.connect(clearCache);
                             }
                         }
+
+                        ListItem {
+                            objectName: "privacy.clearAllCookies"
+                            ListItemLayout {
+                                title.text: i18n.tr("Clear all cookies")
```
UniversalSuperBox, line 210: This option is sentence cased but the two above it are Title Cased. Shall we pick one or the other?

UniversalSuperBox (changes requested):

src/app/webcontainer/WebappSettingsPage.qml
balcy: ~hmm, is "Clear All Cookies" ok ?~

balcy: On the other hand: the first 4 items are not Title Cased (Only allow ..., Domain blacklist/whitelist, Domain specific settings, Camera & microphone)
maybe it is easier to make it "Clear browsing history" and "Clear cache"
It is used in the question for the user, too.
So I would vote for this.

balcy:
hmm the build problem is strange (adapting the Jenkinsfile seemed to have worked on branch https://github.com/ubports/morph-browser/pull/234

I think we might be able to merge it to xenial like that one: 
https://github.com/ubports/morph-browser/pull/256

UniversalSuperBox:
@balcy, it appears that commits from the xenial branch have appeared behind your commits in this branch. Could you please re-create this PR or push the branch so that only your changes on top of the xenial branch are here?

PR #244, opened 2019-10-18
allow empty homepage

### LINKED ISSUE #236: Get back to a blank home page
To reproduce this bug:
Go to the Settings page
In Home Page, type a website (for example https://ubports.com)
Save
Then reopen Home Page and clear the URL to get back to a blank home page

Bug:
Once the URL is void, it is impossible to click on the Save button

Expected behavior:
The Save button is accessible even when the home page URL is void
In this case, a blank page is defined as the home page

balcy:
fixes https://github.com/ubports/morph-browser/issues/236

### DIFF
--- a/src/app/webbrowser/SettingsPage.qml
+++ b/src/app/webbrowser/SettingsPage.qml
@@ -471,16 +471,6 @@ FocusScope {
             }
 
             Button {
-                objectName: "homepageDialog.cancelButton"
-                anchors {
-                    left: parent.left
-                    right: parent.right
-                }
-                text: i18n.tr("Cancel")
-                onClicked: PopupUtils.close(dialogue)
-            }
-
-            Button {
                 id: saveButton
                 objectName: "homepageDialog.saveButton"
                 anchors {
@@ -503,6 +493,16 @@ FocusScope {
 
                     PopupUtils.close(dialogue);
                 }
+            }
+            
+            Button {
+                objectName: "homepageDialog.cancelButton"
+                anchors {
+                    left: parent.left
+                    right: parent.right
+                }
+                text: i18n.tr("Cancel")
+                onClicked: PopupUtils.close(dialogue)
             }
         }
     }
--- a/src/app/webbrowser/SettingsPage.qml
+++ b/src/app/webbrowser/SettingsPage.qml
@@ -82,21 +82,22 @@ FocusScope {
                     id: homepageListItem
                     objectName: "homepage"
                     readonly property url currentHomepage: settingsObject.homepage
+                    readonly property url defaultHomepage: "https://start.duckduckgo.com"
 
                     ListItemLayout {
                         title.text: i18n.tr("Homepage")
-                        subtitle.text: homepageListItem.currentHomepage
+                        subtitle.text: (homepageListItem.currentHomepage.toString() === "") ? i18n.tr("New Tab Page") : homepageListItem.currentHomepage
                         Icon {
                             id: resetHomepage
                             name: "reset"
 
                             height: units.gu(2)
                             width: height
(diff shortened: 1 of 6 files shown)


UniversalSuperBox (changes requested):
Deleting everything to get the new tab page isn't terribly intuitive... It might be better to provide a "Use New Tab" button on the dialog like other browsers do.

Also make sure that the New Tab page and the "Homepage" menu item in the Settings say "New Tab Page" or something similar, since just a blank entry is a little disconcerting. Not displaying "Homepage" on the New Tab page when there is no homepage might be a good idea.

UniversalSuperBox:
As an aside, do we want to change the default homepage to the new tab page once this is implemented? That might be better for new user?

src/app/webbrowser/SettingsPage.qml
```diff
@@ -86,6 +86,24 @@ FocusScope {
                     ListItemLayout {
                         title.text: i18n.tr("Homepage")
                         subtitle.text: homepageListItem.currentHomepage
+                        Icon {
+                            id: resetHomepage
+                            name: "reset"
+
+                            height: units.gu(2)
+                            width: height
+                            opacity: (settingsObject.homepage.toString() === settingsObject.getDefaultHomepage()) ? 0.5 : 1
```
cibersheep: - Two questions:
 1) Isn't possible to use `color: theme.palette.disabled.backgroundText`?
 2) Does the MouseArea need to be disabled? I know would do nothing the second time but still

src/app/webbrowser/SettingsPage.qml
```diff
@@ -86,6 +86,24 @@ FocusScope {
                     ListItemLayout {
                         title.text: i18n.tr("Homepage")
                         subtitle.text: homepageListItem.currentHomepage
+                        Icon {
```
cibersheep, line 90: @balcy 
- Two questions:
 1) Isn't possible to use `color: theme.palette.disabled.backgroundText`?
 2) Does the MouseArea need to be disabled? I know would do nothing the second time but still

UniversalSuperBox: Hmm, there's no feedback when you tap the icon, is there? Like a darker background until you release

UniversalSuperBox: Aha, that'd be the use of an AbstractButton here

balcy: @cibersheep 
1) you mean instead of the opacity ?
I've seen that at several other locations in the code the opacity of 0.3 is used for inactive items, I could use this value instead.
2) if I disable it, the user can click through it and the homepage dialog opens, which should IMO only happen on the left side.
@UniversalSuperBox  I've tried with AbstractButton (and ChromeButton) but without luck. (the positioning did not work)

balcy: e.g. in https://github.com/ubports/morph-browser/blob/xenial/src/app/ChromeButton.qml 1.0 / 0.3 is used

cibersheep: > @cibersheep
>     1. you mean instead of the opacity ?
>        I've seen that at several other locations in the code the opacity of 0.3 is used for inactive items, I could use this value instead.

Fair enough :)
 
>     2. if I disable it, the user can click through it and the homepage dialog opens, which should IMO only happen on the left side.

I see, the tap gets detected by the ListItem. Yep. That would be more confusing

UniversalSuperBox (changes requested):
image

Remember, main action always on top

balcy:
> Remember, main action always on top

blame original webbrowser-app : )
https://github.com/ubports/webbrowser-app/blob/85bba9e7c41d29dba30658103e59ffb2ef579b52/src/app/webbrowser/SettingsPage.qml#L330

UniversalSuperBox (approved):

PR #250, opened 2019-11-01
Remove the DDG search with canonical string

Aury88:
Removed the DuckDuckGo search engine that adds the "t=canonical" string to the searches. this could lead to misunderstanding or (wrong) inference like this https://forums.ubports.com/topic/3309/morph-browser-explanation-about-canonical-in-url
also ubports is indipendent from canonical so this could lead to wrong statistic analisys about ubuntu diffusion (with accusations of inflating data) and biases about ubports project and its indipendence. use instead the duckduckgo-html-only.xml (renamed duckduckgo.xml)


balcy:
yeah I've noticed that "canonical" too.
But why not keeping both duckduckgo (without the t=canonical) and "duckduckgo html only" ?

UniversalSuperBox (changes requested):
The search engine can stay, but removing the 'canonical' tag should be fine.

Aury88:
see commit #251(don't know how to revert in a PR)

PR #251, opened 2019-11-01
remove only the t=canonical string

Aury88:
Revert commit 04f61fcc88d06137525c7866fc0e5029dffab266 that added the t=canonical string to the searches.
substitute the "Remove the DDG search with canonical string #250"


UniversalSuperBox (approved):
Once this builds, it's good to go. Tried it out on my Nexus 5 just now.

Issue #107, opened 2018-11-20
Excessively zoomed-in page(s)


jezek:
Hello, there is at least one site, that is excessively zoomed in and can't be zoomed out.
This is very similiar to issue #12  (or perhaps the same), which is closed as solved.
The old browser in 15.04 (Oxide) got the scaling right. So, I think, it's a regression bug.

Device: FP2
OS: Ubuntu 16.04 (2018-W46)
Site: [memrise.com](https://www.memrise.com/home/)
![screenshot20181114_180617523 resized 25](https://user-images.githubusercontent.com/5408235/48787230-af27f780-ece8-11e8-9d7c-5ae43374b227.jpg) ![screenshot20181114_180658246 resized 25](https://user-images.githubusercontent.com/5408235/48787232-afc08e00-ece8-11e8-8135-c9f249e3193f.jpg)

Maybe, there are more sites experiencing this zoom bug out there.

This topic is being discussed on [this ubports forum](https://forums.ubports.com/topic/1883/new-browser-excessively-zoomed-in).

halucigenia:
I have the same issue on https://www.three.co.uk/New_My3/My3_Home
It's a pain as I can't top up from there on my phone.

casimir7:
I also have this issue frequently (FP2, OTA-8). 
**One idea on this:**
Most websites choose their representation according to the browser identification code sent.
If you look at [http://sqa.fyicenter.com], it shows:

Device type:
Operating system: Linux
Mozilla/5.0 (Linux; Ubuntu 16.04 like Android 4.4) AppleWebKit/537.36 Chrome/65.0.3325.151 Mobile Safari/537.36)

On an iphone it shows:
Device type: Mobile Phone

It might thus be that some sites do not recognize the device as mobile and display a default Desktop version. 
Anyway, it would be good (as a generic solution) to have the zoom not only to work in positive direction but also in negative direction ALWAYS, even at 100%.

zubozrout:
Hm, I am not sure what's the best way to approach this, but what about this?

`/morph-browser/src/app/WebViewImpl.qml`
[WebViewImpl.qml.txt](https://github.com/ubports/morph-browser/files/3704489/WebViewImpl.qml.txt)

It seems to work so maybe someone could have a look, try to think of a better solution / tweak this one and merge it :).

It can be downloaded and tested here: https://www.iubuntu.cz/apps/morph-browser-autozoom.click

[...]

PR #255, opened 2019-11-06
Fit to width in zoom controls

### LINKED ISSUE #233: Morph-browser crash when clicking on Settings
Nexus5
Ubports version: 2019-W41

$QTWEBENGINE_CHROMIUM_FLAGS="--enable-logging=stderr --v=1" morph-browser --desktop_file_hint=/usr/share/applications/morph-browser.desktop
Loading module: 'libubuntu_application_api_touch_mirclient.so.3.0.0'
UCUriHandler: Empty "APP_ID" environment variable, ignoring.
unity::action::ActionManager::ActionManager(QObject*):
	Could not determine application identifier. HUD will not work properly.
	Provide your application identifier in $APP_ID environment variable.
error calling result "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.126\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"com.canonical.libertine.Service.Operations\" member=\"list\" error name=\"(unset)\" requested_reply=\"0\" destination=\"com.canonical.libertine.Service\" (uid=32011 pid=3920 comm=\"/usr/bin/python3 /usr/bin/libertined \")"
error calling result "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.126\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"com.canonical.libertine.Service.OperationsMonitor\" member=\"running\" error name=\"(unset)\" requested_reply=\"0\" destination=\"com.canonical.libertine.Service\" (uid=32011 pid=3920 comm=\"/usr/bin/python3 /usr/bin/libertined \")"
lastError - no arguments?
error calling result "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.126\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"com.canonical.libertine.Service.OperationsMonitor\" member=\"last_error\" error name=\"(unset)\" requested_reply=\"0\" destination=\"com.canonical.libertine.Service\" (uid=32011 pid=3920 comm=\"/usr/bin/python3 /usr/bin/libertined \")"
lastError - no arguments?
error calling result "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.126\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"com.canonical.libertine.Service.OperationsMonitor\" member=\"result\" error name=\"(unset)\" requested_reply=\"0\" destination=\"com.canonical.libertine.Service\" (uid=32011 pid=3920 comm=\"/usr/bin/python3 /usr/bin/libertined \")"
lastError - no arguments?
WebEngine compiled with X11 support, however qpa backend is not xcb. This may fail.
Sandboxing disabled by user.
[6895:6895:1010/224519.842092:VERBOSE1:zygote_main_linux.cc(588)] ZygoteMain: initializing 0 fork delegates
[6895:6895:1010/224519.843056:INFO:cpu_info.cc(50)] Available number of cores: 4
[6877:6877:1010/224519.879958:VERBOSE1:webrtc_internals.cc(110)] Could not get the download directory.
file:///usr/share/morph-browser/webbrowser/Toolbar.qml:40: ReferenceError: bug is not defined
file:///usr/share/morph-browser/webbrowser/Toolbar.qml:39: ReferenceError: bug is not defined
file:///usr/share/morph-browser/webbrowser/Browser.qml:1003:33: Unable to assign double to QQuickAnchorLine
library "/vendor/lib/egl/libGLESv2S3D_adreno.so" not found
This plugin does not support raise()
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/MorphWebView.qml:107:5: QML Connections: Cannot assign to non-existent property "onContextMenuIntercepted"
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/MorphWebView.qml:107:5: QML Connections: Cannot assign to non-existent property "onInsertionHandleTapped"
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/MorphWebView.qml:107:5: QML Connections: Cannot assign to non-existent property "onStatusChanged"
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/MorphWebView.qml:108:17: Unable to assign [undefined] to QObject*
file:///usr/share/morph-browser/ScrollTracker.qml:54:5: QML Connections: Cannot assign to non-existent property "onContentYChanged"
APP_ID isn't set, the handler can not be registered
[6877:6906:1010/224520.288064:ERROR:zygote_host_impl_linux.cc(271)] Failed to adjust OOM score of renderer with pid 6922: Permission denied (13)
[6877:6908:1010/224520.327726:ERROR:nss_util.cc(653)] Error initializing NSS with a persistent database (sql:/home/phablet/.pki/nssdb): NSS error code: -8126
[6877:6908:1010/224520.327836:VERBOSE1:nss_util.cc(659)] Initializing NSS without a persistent database.
[6877:6908:1010/224520.331839:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Argon2018' log
[6877:6908:1010/224520.331924:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Argon2019' log
[6877:6908:1010/224520.333398:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Argon2020' log
[6877:6908:1010/224520.333495:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Argon2021' log
[6877:6908:1010/224520.333554:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Aviator' log
[6877:6908:1010/224520.333611:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Icarus' log
[6877:6908:1010/224520.333669:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Pilot' log
[6877:6908:1010/224520.333724:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Rocketeer' log
[6877:6908:1010/224520.333831:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Google 'Skydiver' log
[6877:6908:1010/224520.333892:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Cloudflare 'Nimbus2018' Log
[6877:6908:1010/224520.333948:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Cloudflare 'Nimbus2019' Log
[6877:6908:1010/224520.334008:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Cloudflare 'Nimbus2020' Log
[6877:6908:1010/224520.334066:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Cloudflare 'Nimbus2021' Log
[6877:6908:1010/224520.334131:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: DigiCert Log Server
[6877:6908:1010/224520.334186:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: DigiCert Log Server 2
[6877:6908:1010/224520.334245:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Symantec log
[6877:6908:1010/224520.334299:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Symantec 'Vega' log
[6877:6908:1010/224520.334353:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Symantec 'Sirius' log
[6877:6908:1010/224520.334407:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: WoSign log
[6877:6908:1010/224520.334461:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Venafi Gen2 CT log
[6877:6908:1010/224520.334516:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: CNNIC CT log
[6877:6908:1010/224520.334571:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: StartCom log
[6877:6908:1010/224520.334626:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Comodo 'Sabre' CT log
[6877:6908:1010/224520.334680:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Comodo 'Mammoth' CT log
[6877:6908:1010/224520.334754:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Izenpe log
[6877:6908:1010/224520.334828:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Venafi log
[6877:6908:1010/224520.334886:VERBOSE1:multi_log_ct_verifier.cc(75)] Adding CT log: Certly.IO log
qml: Loaded 14 UA override(s) from file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/ua-overrides-mobile.js
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Morph/Web/MorphWebContext.qml:100: Error: Invalid write to global property "userAgentOverrides"
APP_ID isn't set, the handler ignored
[6877:6908:1010/224520.838661:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/pt_PT/blog/ubports-blog-1/post/ubuntu-touch-ota-8-release-207
qml: user agent: Mozilla/5.0 (Linux; Ubuntu 16.04 like Android 4.4) AppleWebKit/537.36 Chrome/65.0.3325.151 Mobile Safari/537.36
[6877:6908:1010/224524.516556:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: blob:https://ubports.com/47423a59-1b44-439f-9411-fb473c02446a
[6877:6908:1010/224524.565807:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/content/2279-399f038/web.assets_common.0.css
[6877:6908:1010/224524.569546:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/content/6115-d8469fa/web.assets_frontend.0.css
[6877:6908:1010/224524.572877:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/content/6116-d8469fa/web.assets_frontend.1.css
[6877:6908:1010/224524.578794:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/res.company/1/logo?unique=1d3573e
[6877:6908:1010/224524.580648:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/5321/screenshot20190304_100506825.png
[6877:6908:1010/224524.582660:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/5322/screenshot20190304_103023615.png
[6877:6908:1010/224524.587257:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/blog.post/206/author_avatar?unique=63ed260
[6877:6908:1010/224524.589435:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/3748/2008a.png
[6922:6922:1010/224524.606252:VERBOSE1:LocalWindowProxy.cpp(228)] A context is created NOT from snapshot
[6922:6922:1010/224524.627251:VERBOSE1:LocalWindowProxy.cpp(228)] A context is created NOT from snapshot
[6877:6908:1010/224524.661189:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/content/2284-399f038/web.assets_common.js
[6877:6908:1010/224524.661890:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/content/6118-d8469fa/web.assets_frontend.js
[6877:6908:1010/224524.930784:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/5293?access_token=9e380443-9a5c-4114-98e0-3769ee31c4b8
[6877:6908:1010/224524.939240:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/5320?access_token=418b1323-3bb7-4f64-afb7-5ec8fdc8140a
[6877:6908:1010/224524.945913:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/static/lib/fontawesome/fonts/fontawesome-webfont.woff2?v=4.7.0
[6877:6908:1010/224525.159369:VERBOSE1:loader_util.cc(45)] To buffer: https://ubports.com/web/static/lib/fontawesome/fonts/fontawesome-webfont.woff2?v=4.7.0
[PERFORMANCE]: Last frame took 65 ms to render.
[PERFORMANCE]: Last frame took 59 ms to render.
[PERFORMANCE]: Last frame took 48 ms to render.
[6877:6908:1010/224526.453350:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/website/translations
[6877:6908:1010/224526.492672:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/webclient/locale/pt_PT
[6877:6908:1010/224526.639531:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://analytics.ubports.com/piwik.js
qml: [JS] (https://ubports.com/web/content/6118-d8469fa/web.assets_frontend.js:1772) DOM doesn't contain '.o_payment_form'
[6877:6877:1010/224526.706942:INFO:CONSOLE(1772)] "DOM doesn't contain '.o_payment_form'", source: https://ubports.com/web/content/6118-d8469fa/web.assets_frontend.js (1772)
[6922:6922:1010/224526.747181:VERBOSE1:ThreadState.cpp(1312)] [state:0xb6e17198] CollectGarbage: time: 26ms stack: NoHeapPointersOnStack type: GCWithoutSweep reason: IdleGC
[6922:6922:1010/224526.754881:VERBOSE1:ThreadState.cpp(980)] [state:0xb6e17198] PostSweep: collection_rate: 32%
[PERFORMANCE]: Last frame took 40 ms to render.
[6877:6908:1010/224527.006222:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://analytics.ubports.com/piwik.php?action_name=Ubuntu%20Touch%20OTA-8%20Release%20%7C%20UBports&idsite=1&rec=1&r=559090&h=22&m=45&s=26&url=https%3A%2F%2Fubports.com%2Fpt_PT%2Fblog%2Fubports-blog-1%2Fpost%2Fubuntu-touch-ota-8-release-207&_id=3d5715ead3ffbe76&_idts=1543182791&_idvc=17&_idn=0&_refts=1570740248&_viewts=1570740248&_ref=https%3A%2F%2Fduckduckgo.com%2F&send_image=1&cookie=1&res=432x768&gt_ms=3388&pv_id=tpw2yD
[6877:6908:1010/224527.080613:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/website/static/src/xml/website.gallery.xml
[6877:6908:1010/224527.279599:VERBOSE1:loader_util.cc(45)] To buffer: https://ubports.com/website/static/src/xml/website.gallery.xml
[PERFORMANCE]: Last frame took 52 ms to render.
[6877:6908:1010/224527.372587:VERBOSE1:network_delegate.cc(30)] NetworkDelegate::NotifyBeforeURLRequest: https://ubports.com/web/image/website/1/favicon/
qml: [JS] (https://ubports.com/web/content/6118-d8469fa/web.assets_frontend.js:1624) Uncaught TypeError: Cannot read property 'dataset' of undefined
[6877:6877:1010/224528.706779:INFO:CONSOLE(1624)] "Uncaught TypeError: Cannot read property 'dataset' of undefined", source: https://ubports.com/web/content/6118-d8469fa/web.assets_frontend.js (1624)
[PERFORMANCE]: Last frame took 32 ms to render.
propsReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.129\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"org.freedesktop.DBus.Properties\" member=\"GetAll\" error name=\"(unset)\" requested_reply=\"0\" destination=\"org.freedesktop.NetworkManager\" (uid=0 pid=2154 comm=\"NetworkManager \")"
nmReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.129\" (uid=32011 pid=6877 comm=\"morph-browser --desktop_file_hint=/usr/share/appli\") interface=\"org.freedesktop.NetworkManager\" member=\"GetDevices\" error name=\"(unset)\" requested_reply=\"0\" destination=\"org.freedesktop.NetworkManager\" (uid=0 pid=2154 comm=\"NetworkManager \")"
"Object path cannot be empty"
Received signal 11 SEGV_MAPERR 000000000004
#0 0x0000aafb4724 <unknown>
#1 0x0000aa19874c <unknown>
#2 0x0000aafb4aea <unknown>
#3 0x0000b5b37270 <unknown>
[end of stack trace]
Calling _exit(1). Core file will not be generated.

jezek:
Hello, I'm presenting some changes to zoom controls.

**Why?**
A workaround for issue #107.
fixes https://github.com/ubports/morph-browser/issues/233

**What's new?**
- Fit to width button in zoom menu - Zooms page to fit width, using body.scrollWidth.
- Automatic fit to width checkbox in settings - If enabled and page has no specific zoom for domain, fit to width is triggered automatically on first page load for domain, on width or relevant browser settings changed or reset button in zoom menu is pressed.

**What changed?**
- Save button in zoom menu asks where to save. To domain settings or global settings.
- Specific zoom for domain can be saved even if it is the same as default zoom - To prevent triggering autozoom for certain domains.
- After global/default zoom factor is saved/changed, equal domain zoom values are not reset - I think it's bad habbit to touch old settings. This way, you could accidentaly clear all domain zoom settings just by playing with the default zoom factor slider in settings.
- Zoom menu is not hidden on tab surfing - I think, it is more consistent this way. You show it, you hide it!
- Zoom control code was separated to standalone file and refactored.


jezek:
Why this shows failed checks? It is a jenkins bug, or my fault? The app can be build and run by clickable (my version 6.4.0).

balcy:
I've already tested it a bit, and think it's a good improvement (both regarding feature and code). Thank you for the solid work !

I think @cibersheep will find sth. (non-suru icon :) )

one little thing I have noticed:
If you click on "Fit(116%)" as an example, and then "Save"->"Save as default", the default zoom is 116%. 
If you then open the Settings, the zoom is rounded to 115%. I have caused this by:
```
                            onValueChanged: {
                                // round for 5% steps (e.g. 95%, 100%)
                                var percentValue = Math.round(value * 100 / 5) * 5
                                settingsObject.zoomFactor = percentValue / 100
                            }
```
what we could do, is already round the value when saving the default value.  Without that rounding it is quite difficult to select a value with the slider.

a note for the current version:
The reason for clearing the entries with the default zoom factor automatically was, that in my version the user never sets the domain zoom factor explicitly (the entries are created automatically every time the user changes the current zoom for a page). I wanted to clean up as many of them as possible automatically.
So my thinking was: If the user visits some sites, and sets each of them to 110%, and later saves this as the default zoom, all future sites with 110% are no longer saved.
Therefore the old entries with 110% (which is now the default) have become obsolete, too.

But I understand your explanation too, and for the new version it does no longer make sense. And also generally you are right that the user should not loose settings accidently, so 👍.

Another question: Even if I have not activated the "Automatic Fit to Width", after each page load the fit factor is calculated. Would it be possible to only calculate the Fit after the user presses the button, or only if the zoom menu is open ? I think otherwise it would mostly calculating it without the need for the user.

cibersheep:
Have you called the Color Police (aaaand icon)? XD
What do you need? Tell me, tell me

balcy:
dialog with 2 positive actions detected :)
I think the save dialog does contain all needed information. Not sure if it needs adaptions.

cibersheep:
Ouh. That's why all alarms went off :D 
Just choose the main main action (the one you consider the most used) and color that accordantly ;)

I can't test in some hours / day

balcy:
I think there is no case with both "Save as default" and "Save for Domain" buttons are enabled at the same time.
Therefore we could make 2 dialogs: One for saving the default zoom factor, and one for saving the (default) zoom factor for the current domain.
(Or one dialog with adapted title and visibility of the save button, so that there is only one visible)

jezek:
@balcy 
> If you click on "Fit(116%)" as an example, and then "Save"->"Save as default", the default zoom is 116%.
> If you then open the Settings, the zoom is rounded to 115%....

Ah, that slipped my attention. Somewhere on my path I rounded the fit value to nearest 5 downwards (to not overzoom), but then I decided not to do that. Why? Becouse of duckduckgo. When I did a search on my phone (FP2), the fit to width factor is calculated to 93%. If rounded, it would be 90% and the auto zoom would kick in. I personally found that annoying. So I decided not to round the zoom and autozoom only if fit factor difference from current is more or equal 10 percent points. But the rounding is definitely doable. Or we can not round in settings and tweak the input method. Eg change the slider for text input, or add a -/+ buttons arround the slider to decrease/increase by 1 to fine tune. I have some time I could do any variant. I don't know which solution is better. But if we decide to not round, I had an idea. To write the step size under zoom in/out button in zoom menu (curently -10/+10) and by long press on one of the buttons the step size will cycle around some predefined step sizes (+/-1, +/-5, +/-10)


> The reason for clearing the entries...

I fully understood the reason and that now (with domain zoom saves) the clearing is obsolete, that's why I didn't hesitate to make this change ;)


> Another question: Even if I have not activated the "Automatic Fit to Width", after each page load the fit factor is calculated. Would it be possible to only calculate the Fit after the user presses the button, or only if the zoom menu is open ? I think otherwise it would mostly calculating it without the need for the user.

2 or 3 commits ago, It was done that way (only if zoom menu is visible). But then at some point in refactoring and rewriting I decided to calculate the fit factor every time, cause I think it is not so resource hungry (one js variable extracion & a few conditions) and it is done once per domain (unless you change width/rotation, or play with settings often) so I decided to calculate even when zoom menu is not visible. But, I can bring that calculate only if zoom menu is visible logic easily back if really desired. (note that fit calculation does not happen unless webview is visible, so if many tabs are open, only for the visible one fit is calculated after width or settings change).


@cibersheep I think first the alarm went of cause I use a zoom-fit-best icon for a "fit to width" button, which is not in the suru set. However the icon is in icon folders in phone, but not if you run as ```clickable desktop```. I couldn't find out which other "legal" icon to use. Can you help with that please?

@balcy @cibersheep 
> dialog with 2 positive actions detected :)
> I think the save dialog does contain all needed information. Not sure if it needs adaptions.
> ... 
> Just choose the main main action (the one you consider the most used) and color that accordantly ;)

If there is no other way and I had to decide which one to choose as main action, I would pick the save domain action.
But. What if I can't decide which action is most used? What if, theoretically, the actions are equally important? What coloring will it be? None or some other? Cause I wanted it make different as cancel.

To clarify. There is a popup menu with 3 options. ```Save to domain```, ```Save as default``` and ```Cancel```. How to properly color if the two save options are equal? Also what is the proper color for cancel button (it just closes the popup).

jezek:
@balcy 
> I think there is no case with both "Save as default" and "Save for Domain" buttons are enabled at the same time.

Yes there is a case. If you have turned "autozoom" on and visit a wrong zoomed page (eg. pedf.cuni.cz). The current zoom (autofitted) will be other than default, so it can be saved as default zoom and also can be saved as zoom for domain ;)

balcy:
oops sorry, I have only tested without auto zoom ~on~ off (normal and private pages).

Hmm two save buttons are probably too much. It could be ok to make the "Save for domain" button without the positive color then.

The other question: I think cancel buttons have just the default color (no color set)

balcy:
[correction: off]

jezek:
> Hmm two save buttons are probably too much. It could be ok to make the "Save for domain" button without the positive color then.
> [correction: off]

Just for clarification. I understood, that only ```Save for domain``` will be positive/green if both are enabled and all other will be default. If only one is enabled, then the enabled one will be positive. Is it right?

EDIT: What about the blue color button. I've seen it somewhere. Can we use that?

balcy:
I think only one green button is possible, and the rest must not have colors, but let's wait for @cibersheep

jezek:
I see, they mention a neutral color in [guidelines](https://gitlab.com/ubports/teams/ux-des/ubuntu-touch-human-interface-guidelines---hib/wikis/General%20Concepts/01-System-Palette). So maybe one of them (or both) can be neutral. But yes, wait for Joan's opinion.

What about the fit to width rounding and computation problems? Which way to go? It's on us to decide, or is here some committee to decide for us?

cibersheep:
> @cibersheep I think first the alarm went of cause I use a zoom-fit-best icon for a "fit to width" button, which is not in the suru set. However the icon is in icon folders in phone, but not if you run as `clickable desktop`. I couldn't find out which other "legal" icon to use. Can you help with that please?

I see. We have no analog icon in Suru. I can create one tomorrow


> @balcy @cibersheep
> 
> To clarify. There is a popup menu with 3 options. `Save to domain`, `Save as default` and `Cancel`. How to properly color if the two save options are equal? Also what is the proper color for cancel button (it just closes the popup).

Don't worry if the other action is the same grey as cancel (look at power down menu).
I'd go with just coloring the top button ;)

balcy:
about the rounding: I think I have an idea how to keep your code and solve it by a small fix in the SettingsPage / WebApp SettingsPage. I will write you later today.

balcy:
ok I've included it here:
https://github.com/jezek/morph-browser/pull/1

balcy:
all the morph branches seem to have trouble with building right now 
`ERROR: some files failed to be added` in the "Results" part.
@UniversalSuperBox any problem with the build server ?

UniversalSuperBox:
Oh, sorry. A change to the build script requires an update to the Jenkinsfile, like https://github.com/ubports/messaging-app/commit/51d303d0a1c692cbb7cf6040acde4daca1bc2a35

We're working on making the standard Jenkinsfile an import so that we don't need to do this ever again, just change the Jenkinsfile in one place and it's done. For now, though... that's the fix.

jezek:
@balcy Your fixes were implemented. I took liberty to change the condition, to improve readability (no need to use the Math.abs in this case) and added comments.

In zoom save dialog, I tried the positive button color on both separately. The top button green felt more visual pleasing, so the top button is positive. If you insist, to be positive the ```Save as default``` button, I can switch their position, to stay the top button positive.

Also the code for fetching fit to width factor only if zoom menu visible (for disabled autofit) is back. So no more body.scrollWidth fetching until really needed.

@UniversalSuperBox The jenkins file was modified as instructed, but checks failed again. There is some error in build-repo.sh. Also the jenkins file seemed already modified, cause the changes were only cosmetic (see https://github.com/ubports/morph-browser/pull/255/commits/9405ee84132f370d8be3e5919c5ced125299b6c4).

balcy (dismissed):

balcy:
hmm the build problem is strange (adapting the Jenkinsfile seemed to have worked on branch https://github.com/ubports/morph-browser/pull/234

I think we might be able to merge it to xenial like that one: 
https://github.com/ubports/morph-browser/pull/256

balcy (dismissed):

UniversalSuperBox (changes requested):
"Zoom to fit" behaves very strangely for me on my Fairphone 2. For example, head to github.com to get to their "Built for developers" homepage. Then go to the zoom menu. For me, "Fit" starts at 101% or 102%. If I select this, the zoom to fit may be recalculated to somewhere around 112% or it may stay grayed out. Generally it will be at "zoom+1" for whatever zoom level I'm at.

I find the "Save" option confusing. There is already a place in the settings to change the default zoom, do we need it in another place? More importantly, do we need to crowd this already-crowded menu any more?

image

cibersheep:
Aaaand now I see the non-Suru icon :D
Tomorrow I'll have one done

cibersheep:
> I find the "Save" option confusing. There is already a place in the settings to change the default zoom, do we need it in another place? More importantly, do we need to crowd this already-crowded menu any more?
> 

Morph is getting really busy menues.
- What about saving automatically the zoom level?
- And tapping outside of the popup to close (as copy/paste popup)?

jezek:
@UniversalSuperBox 
> "Zoom to fit" behaves very strangely for me on my Fairphone 2. For example, head to github.com to get to their "Built for developers" homepage. Then go to the zoom menu. For me, "Fit" starts at 101% or 102%. If I select this, the zoom to fit may be recalculated to somewhere around 112% or it may stay grayed out. Generally it will be at "zoom+1" for whatever zoom level I'm at.

Yes, this is known behaviour. The ```fit to width``` factor is computed from ```body.scrollWidth``` of current page. Most pages have responsive design. Some of them get it right and the ```fit to width``` factor is the same as current zoom, some will always have the factor slightly off. This can't be helped and is a huge pain. For ```automatic fit to width```, that behavior caused, what I call "autofit zoom traveling", where the zoom was adjusted more times and sometimes would end up in max/min value. For ```automatic fit to width``` this was solved by reading the ```body.scrollWidth``` value only at ```default zoom factor```, which causes a visible zoom page jumps after some actions (reset, width change, ...). I didn't want to do these zoom jumps if ```automatic fit to width``` is disabled (by default). 

My suggestions are:
1. Leave it as it is - There will be some confusion, but it's better than nothing, and you can easily adjust zoom with one tap if needed.
2. Remove the ```fit to width``` button - It's confusing, nobody needed it before, it's better to not have this button.
3. Remove the factor number from ```fit to width``` button - No confusion, button will be always enabled and the "zoom traveling" after the button is repeatedly pressed can be solved as in autozoom.

> I find the "Save" option confusing. There is already a place in the settings to change the default zoom, do we need it in another place? More importantly, do we need to crowd this already-crowded menu any more?

The ```save``` button was already there, I only added an option into the save popup. I think it was the work of @balcy, he has some saying in it.
Suggestions:
1. Leave it as it is - It was there for some time, nobody complained, some people are already used to it, don't take away the button to them.
2. Remove it completely - Default zoom can be set from settings and domain zoom is saved automatically after zoom in/out (and fit to width) buttons pressed.
3. Leave the button there, but let it only do only one action (save domain or default) and perhaps without popup - this would need some discussion.

@cibersheep 
> aaand now I see the non-Suru icon :D
> Tomorrow I'll have one done

Maybe it won't be needed, if we decide to remove the ```fit to width``` button.

> What about saving automatically the zoom level?

The domain zoom level is saved automatically after zoom in/out (and fit to width) buttons pressed.

> And tapping outside of the popup to close (as copy/paste popup)?

Yop, it's doable, but I will wait to see if we decide to ditch the ```save``` button, so there wil be no popup ;)

@balcy @UniversalSuperBox @cibersheep What are your opinions suggestions. I don't really mind if things stay as they are now (except the popup closing). But on the other hand I don't really mind if we ditch the ```fit to width``` and ```save``` buttons. The menu will become more minimalistic (-, reset, +, close) and ```automatic fit to width``` will work anyway (if enabled).

balcy:
I would leave the current state. [edit: current state of the PR]
- the context menu does have a close button, too
it used to be called "Cancel" for a while, but was renamed to "Close" both in context menu and zoom menu:
https://github.com/ubports/morph-browser/pull/89/commits/59c88dd7ce78a88ef7444dc27a9bcf1e5f42615d , see screenshots / discussion here: https://github.com/ubports/morph-browser/pull/89

Right now both of those menus don't take away the focus from the webview, so that you can scroll / select or edit text while they are open.

- the save and reset buttons are useful, too IMO.

for the "Fit to Width" button I don't have a preference, but there seems to be enough space available for it.

balcy:
https://github.com/ubports/morph-browser/pull/89#issuecomment-439630215
what is funny is that the responses to this comment came exactly (!) on Nov 19th 2018 : )

UniversalSuperBox:
If the "Fit to width" button's behavior is unpredictable and we *expect* that, I don't think we should have it. Instead, let's try to figure out if we need to fit to width automatically (which this PR does).

Since zoom is saved automatically and the Reset button exists, I don't think bespoke "Save" and "Cancel" buttons are needed. Indeed, "Cancel" is "Close" now.

Even if space is available, we should use it very sparingly. Too many options kill.

balcy:
but why should a PR for "fit to width" suddenly remove existing buttons like save and close ?

The reason why the "save for current domain" was added to the save action, is explained in the first comment:

>What changed?
> ...
>    Specific zoom for domain can be saved even if it is the same as default zoom - To prevent triggering autozoom for certain domains.

balcy:
the following would be strange for me as a user: 
- in a webapp or morph-browser,  long press on the page, the context menu opens, it has a "close" button (it cannot be closed by clicking somewhere on the page)
- click on "zoom" -> the zoom menu opens and the context menu closes
- the zoom menu does not have a close button (it can only be closed by clicking somewhere on the page)
so the context menu and the zoom menu belong together. And changing the context menu with this PR is even more out of its scope.

UniversalSuperBox:
I understand, removing any existing options can wait until later. Let it be known that I meant "Only one button needs to close," since both "Save" and "Close" did the same thing. The menu can have a close button, that's fine.

> Specific zoom for domain can be saved even if it is the same as default zoom - To prevent triggering autozoom for certain domains.

@jezek, did this situation occur often in your testing? Can you give me an example of how I would reproduce this behavior in a way that's confusing for the user?

jezek:
@UniversalSuperBox 
> > Specific zoom for domain can be saved even if it is the same as default zoom - To prevent triggering autozoom for certain domains.
> 
> @jezek, did this situation occur often in your testing? Can you give me an example of how I would reproduce this behavior in a way that's confusing for the user?

No, it did not occur often in my testing. It's an hypothetical situation, when you have ```automatic fit to width``` enabled and don't want to use it for some domains in the future for some reason. So, you can save a ```specific zoom``` factor for domain, even if it is a ```default zoom``` factor.

I think, i know why you are asking. I think, you are trying to determine, if the changes in the ```save``` button (the added ```Save to domain``` button in the popup) are necessary. Am I right? The changes were not really necessary, you can save ```specific zoom``` for domain by hitting the +/-/fit buttons, but my thoughts where, that if there is a ```save``` button, why would it only save default zoom, why not both?

As for your other comment.
>If the "Fit to width" button's behavior is unpredictable and we expect that, I don't think we should have it. Instead, let's try to figure out if we need to fit to width automatically (which this PR does).

The main reason for this PR is the ```automatic fit to width``` ability! The qt web engine is a little bit buggy and some pages are excessively zoomed on phone (see #107). Although on chromium browser with the same dimensions (e.g. in developer mode) the pages are properly zoomed. I think of this PR as a workaround for that bug. The ```fit to width``` button just emerged from the chaos of creation. ;)

I also think, i  have a solution to the confusing and superfluous ```fit to width``` button. How about it would "merge" with the ```reset``` button. So if the ```reset``` button is disabled, it would become a ```fit to width``` button. After hitting ```fit to width``` or make some zoom in/out, the button will become reset again, to reset the zoom changes to default. This will ensure, that the ```fit to width``` will be done only from ```default zoom```, so the fit percentage will be always the same and there will be no more confusion. This would make the zoom menu less crowded as it was before, the ```fit to width``` button will be preserved and no more confusing and @cibersheep gets to make a new suru icon. :) A win-win-win situation. How does it sound?

jezek:
I've chatted with @UniversalSuperBox and he pointed out, that the excessive zoom bug is not present in [Demo browser](https://open-store.io/app/demobr.mateo-salta), which uses vanilla QtWebEngine WebView, so it has to be a bug in morph.

If we can figure out the bug, then there will be no need for this PR, which is a workaround...

So I have decided, to try to investigate and fix the bug properly. And after the bug is fixed we will see if there is some need for some changes in this PR. Until then this workaround will not be merged.

Note to bug:
> So, we have both QT_SCALE_FACTOR and the Ubuntu Touch scaling set. I believe both of these are used for different purposes between QWE and Morph code

EDIT: Found out, that there are pages, that have the excessive zoom bug even in demo browser (e.g. pedf.cuni.cz, memrise.com). In chromium's developer mode phone emulation mode these pages are ok.

balcy (commented):
I personally would go with the previous version, or completely without the "fit to width" button.
I don't think it is better for the user (or code), if the buttons appear / disappear like that.
[Edit]
With "clickable desktop" (duckduckgo homepage) I've set the default zoom to 115% in the settings menu, and then opened the zoom menu on the page.
If you then only click on the second item several times, it will always toggle between 114% and 115%
image
That's why I think "Fit" should be an extra button, or omitted.

UniversalSuperBox:
Right, so the "Default fit" now is "Fit to width," which is not necessarily the default zoom.

UniversalSuperBox:
Responding to your edit @balcy, yes, this appears to be the same issue as when the fit button is separate, its behavior is not consistent on "responsive" websites.

However, I've talked with Jezek about this. We've found that the sites that don't work properly in Morph *also* don't work properly in a stock Android browser. 

For example, [this memrise link](https://www.memrise.com/course/2021573/french-1-145/1/) doesn't flow correctly, you can't use horizontal scroll (they disabled the scroll bar), and you can't zoom out through any method but browser zoom. Same on Morph, the Lineage 15.1 and 16 browsers, and Firefox developer mode.

Another example, [Univerzita Karlova](https://cuni.cz/UK-1.html), can show a similar issue. In this case, the page doesn't flow correctly *but* horizontal scroll works. On Android, you can use pinch zoom to see the whole site. You can't do that in Morph.

The interesting thing there is that **we can't use pinch-to-zoom to zoom out further than whatever Morph originally chooses**. Even so, pinch-to-zoom might not be a complete solution... It's not the easiest for people with motor difficulty.

What makes me so concerned about this change is its inconsistency on responsive websites. It may choose a "fit" zoom 1% higher than your set zoom, it might pick your set zoom, or it might pick another value entirely. 

Also, I don't know how much we should work around sites that are just plain broken, even under other mobile browsers. Fit-to-width seems to be a lot to carry to cover other people's mistakes... even if those mistakes do make using the web annoying.

jezek:
@balcy Yes, I see your point. A button should do only one type of thing. Go it. It was not a good idea to merge them.

The problem with separate button is, as @UniversalSuperBox stated, it's incosistency on responsive sites. I agree with him in this point. To make the button consistent would introduce other inconsistencies or unpleasing visual behaviour (I don't say it's impossible, but with my current skills and knowledge I don't currently know how to 100% solve it).

So the button has to go away.

giiba:
As a user, I'd like the website to either be zoomed to fit automatically, or be able to pinch to zoom out like other mobile browsers let me.

I still regularily load a site, start pinching out of motor reflex, then remember I need a menu... which doesn't reliably pop up with zoom options as it is very easy to select some text and bring up the clipboard options. Then I have to long press in random areas of the site until the zoom option is present, and then I can start trying to zoom out.

This auto zoom idea might be smoothing over other's bad coding, but making the user experience better in Morph is only a positive. It sounds - if I follow the discussion - that the zoom factor should be calculated once and then stored so that the calculated zoom doesn't shift about. The bouncing between 114% and 115% due to rounding issues makes the feature seem unpolished from the user perspective.

But figuring out why Morph doesn't let us zoom out would at least put Morph at parity.

balcy (dismissed):

jezek:
OK, the fit button is gone together with resulting unused code.
Any other adjustments?

src/app/webbrowser/SettingsPage.qml
```diff
@@ -129,6 +129,26 @@ FocusScope {
                     }
                 }
 
+                ListItem {
+                    objectName: "autoFitToWidthEnabled"
+
+                    ListItemLayout {
+                        title.text: i18n.tr("Automatic Fit to Width")
+                        subtitle.text: i18n.tr("Zooms to body.scrollWidth after domain changed, if no specific zoom is saved.")
```
UniversalSuperBox: This flows outside of the page, but also doesn't make much sense to and end-user... Maybe something more like "Zoom out to show more content on some websites." @cibersheep, language thoughts?

src/app/webcontainer/WebappSettingsPage.qml
```diff
@@ -49,6 +49,26 @@ FocusScope {
 
                 width: parent.width
 
+                ListItem {
+                    objectName: "autoFitToWidthEnabled"
+
+                    ListItemLayout {
+                        title.text: i18n.tr("Automatic Fit to Width")
```
UniversalSuperBox: Other items on this page are sentence cased, not Title Cased. Please follow that convention.

UniversalSuperBox (changes requested):
Functionally, everything seems quite alright now. We do need to adjust the wording of this settings page, though.

src/app/webbrowser/SettingsPage.qml
cibersheep: Mmm... what about?

"Adjust the width of the website to the window"

cibersheep:
You might find this icon useful :D and Suru

[zoom-to-fit-width.svg.zip](https://github.com/ubports/morph-browser/files/3906683/zoom-to-fit-width.svg.zip)

Preview:

imatge

src/app/webbrowser/SettingsPage.qml
jezek: Description was changed

src/app/webcontainer/WebappSettingsPage.qml
jezek: Sorry, got distraced by the ```Set Desktop mode``` casing. Changed to sentence cased in both settings pages (browser, container)

jezek:
@cibersheep 
> You might find this icon useful :D and Suru

You are great, but the button is gone, icon is not needed in this PR anymore. Maybe someone will use the icon in other project, when it gets to [Suru icon pack](http://docs.ubports.com/projects/icons/en/latest/). Thank you anyway.

balcy:
@jezek would you mind if I apply the changes https://github.com/ubports/morph-browser/pull/265/files directly into this branch ? I've already tested it in this branch. It is replacing the Slider by a SpinBox (QtQuick Controls)

jezek:
@balcy I don't mind, you're welcome.

cibersheep:
> @cibersheep
> 
> > You might find this icon useful :D and Suru
> 
> You are great, but the button is gone, icon is not needed in this PR anymore. Maybe someone will use the icon in other project, when it gets to [Suru icon pack](http://docs.ubports.com/projects/icons/en/latest/). Thank you anyway.

Do not worry. I wasn't sure so it's better to have it and not using it that the other way around :D 
I'll put it in a safe place

jezek:
@balcy can the ```SpinBox``` be themed? On dark theme it glows white.

balcy:
yes it can (https://doc.qt.io/qt-5/qtquickcontrols2-styles.html) e.g. via 
`QQuickStyle::setStyle("Suru");`
or environment variable:
`phablet@bq:~$ QT_QUICK_CONTROLS_STYLE=Suru morph-browser --desktop_file_hint=/usr/share/applications/morph-browser.desktop`
but then I see a black rectangle with white arrows (left and right), and the current zoom value is also black, so you cannot read it.
If we use this, let's just use the subtitle again for the current zoom value, and not write the zoom value within the spin box

balcy:
so there is still the issue, that the text color of the SpinBox in dark theme is black.
This should be solved "upstream" in https://github.com/ubports/qqc2-suru-style/blob/xenial/qqc2-suru/SpinBox.qml I think.

I tried the following to solve it locally:

```                        
import QtQuick.Controls.Styles 1.4
...
    SpinBox {
    id: defaultZoomFactorSelector
    style: SpinBoxStyle {
       textColor: theme.palette.normal.baseText
   }
```
(https://doc-snapshots.qt.io/qt5-5.9/qml-qtquick-controls-styles-spinboxstyle.html)
but unfortunately I get `module "QtQuick.Controls.Styles" is not installed` in the log
But since the use of the Slider actually makes morph crash when Settings are loaded on some devices, I think it is better even if not visually perfect.

UniversalSuperBox:
Right, so now we should be fine functionally but there are 54 commits in this branch... I'd hate to squash and merge because there are now multiple atomic changes here, but I also don't want to put 54 more commits on the history. What do you think?

cibersheep:
> @balcy can the `SpinBox` be themed? On dark theme it glows white.

Instead of a SpinBox could another element work for you? Like a Tumbler or a Slider?

cibersheep:
> Right, so now we should be fine functionally but there are 54 commits in this branch... I'd hate to squash and merge because there are now multiple atomic changes here, but I also don't want to put 54 more commits on the history. What do you think?

I'll vote for not squashing. Might be important in the future...

balcy:
functionally SpinBox works best I think, I'll open a PR for the suru SpinBox later.

balcy:
https://github.com/ubports/qqc2-suru-style/pull/35
(can be tested on the phone by changing /usr/lib/arm-linux-gnueabihf/qt5/qml/QtQuick/Controls.2/Suru/SpinBox.qml, and remove the corresponding .qmlc file)

jezek:
@UniversalSuperBox What is the disadvantage of many commits? I'm just a basic git user, so I'm asking just out of curiosity. Hmm... if the commit number is high, then how about squash between the comments. Then it becomes 10 commits altogether for this PR and this thread would be consistent.

But if there is no real trade-off of having more commits (just cosmetic), then I'm with @cibersheep, it could be important in the future. At least for learning purposes, to show, that all bigger changes are formed by small incremental steps. ;)

UniversalSuperBox:
Part of the reason for having a well-kept version control system is for its history. This allows me to go back in time, as it were, and see the browser as it was before someone changed something.

The disadvantage in this case is that there are *many* commits and not all of them represent a complete, functional change. There are also reversions inside of this history, such as adding then removing the "fit to width" button.

Ideally, every commit in a history is functional and represents something historically significant (relatively speaking). For example, in this PR, we'd want a history that looks something like this:

```
Add a basic AutoZoom feature
Refactor zoom controller into its own file
Remove "fit to width" button
Change autofit settings text
```

Just as an example, but this keeps every point in the history of this PR that is significant - including the addition and later removal of the "Fit to width" button.

There are also some unrelated changes in this branch that may cause merge conflicts -- such as the Jenkinsfile update and the updates to the SpinBox.

Correcting this history is a little fiddly, but is a *very* good Git skill to have. If you'd like, hit me up in a PM and I'll help you out. We might just end up creating a new PR with these changes, but that will be fine. :D

jezek:
@UniversalSuperBox so I red some docs, tutorials & discussions. I didn't do the squashing on this branch, cause it could mess up this PR.

I've cloned the branch to [xenial_-_autofit_squashed](https://github.com/jezek/morph-browser/commits/xenial_-_autofit_squashed), rebased to current upstream xenial branch and squashed it there. Is it better, or it needs more? (Note: I've changed/squashed only my commits, not Chris's)

Just one question. If the squashed branch (```xenial_-_autofit_squashed```) gets merged and I delete the branch of this PR (```xenial_-_autozoom```), will this discussion references stay in tact? Or should I never delete?

UniversalSuperBox:
@jezek, very nice!

This PR will stay intact either way even if closed, so we can reference here if we need to.

The last thing I'd like to see of this branch is to *not* put all the new changes on top -- that's the PR commits for Weblate or bookmarks. Now that you have the branch squashed, though, you could do something like this to rebase your squashed commits:

```
git checkout xenial
git pull upstream xenial --ff-only  # Assuming the UBports repository is added as a remote called 'upstream'
git checkout -b yet-another-autozoom-branch
git cherry-pick b9eb6c49b1824930d558764c99dc1bfd1c6fe12b..48077636ec2f4e5e5162b867021a6c1b47514fea
```

This checks out and pulls the newest version of the `xenial` branch from this repository, creates another branch based on it, and pulls in your changes *on top of* it. That way, no more commits from other PRs on top of yours!

jezek:
@UniversalSuperBox I've tried your 

> @jezek, very nice!
> 
> This PR will stay intact either way even if closed, so we can reference here if we need to.
> 
> The last thing I'd like to see of this branch is to _not_ put all the new changes on top -- that's the PR commits for Weblate or bookmarks. Now that you have the branch squashed, though, you could do something like this to rebase your squashed commits:
> 
> ```
> git checkout xenial
> git pull upstream xenial --ff-only  # Assuming the UBports repository is added as a remote called 'upstream'
> git checkout -b yet-another-autozoom-branch
> git cherry-pick b9eb6c49b1824930d558764c99dc1bfd1c6fe12b..48077636ec2f4e5e5162b867021a6c1b47514fea
> ```
> 
> This checks out and pulls the newest version of the `xenial` branch from this repository, creates another branch based on it, and pulls in your changes _on top of_ it. That way, no more commits from other PRs on top of yours!

I don't understand. The ```jezek:xenial_-_autofit_squashed``` branch already sits on top of ```ubports:xenial``` branch. The instructions you provided would do the same, as I did (except you got the commit shas in wrong order, your cherry pick command returns empty set error, and if the shas are reversed, the first commit ```4807763``` will be omitted). The result would be the same. Or am I missing something?

UniversalSuperBox:
You know what, I read GitHub in the wrong order. Commits in PRs (such as [this PR's commit log](https://github.com/ubports/morph-browser/pull/255/commits)) are listed from oldest to newest. The normal commit log is listed newest to oldest. Your squashed branch looks good. If you open a PR, let's merge it.

UniversalSuperBox:
Closing in favor of #270

PR #268, opened 2019-11-30
Xenial   bookmarkscorrection

### LINKED ISSUE #266: There are no bookmarks on BQ M10 to use
```
Device: BQ M10
Channel: dev
```
It is no longer possible to use bookmarks on the device BQ M10 in the channel dev.
Clicking on the star icon for bookmarks will bring up the confirmation menu, but the list for bookmarks will remain blank. The bookmark is created, however, because when the address is recalled, the star appears orange in color. However, earlier bookmark categories are available.

balcy:
fixes https://github.com/ubports/morph-browser/issues/266

### DIFF
--- a/src/app/webbrowser/BookmarksFoldersViewWide.qml
+++ b/src/app/webbrowser/BookmarksFoldersViewWide.qml
@@ -159,7 +159,7 @@ FocusScope {
         delegate: DraggableUrlDelegateWide {
             objectName: "bookmarkItem"
 
-            property var entry: folders.currentItem.isAllBookmarksFolder && (homeBookmarkUrl.toString() !== "") ? modelData : model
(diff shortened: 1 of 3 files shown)


src/app/webbrowser/BookmarksFoldersViewWide.qml
```diff
@@ -134,34 +134,34 @@ FocusScope {
...
             }
         }
 
         model: {
             if (!folders.currentItem || !folders.currentItem.model) {
-                return null
+                return null;
             }
 
-            if (folders.currentItem.isAllBookmarksFolder && (bookmarksFoldersViewItem.homeBookmarkUrl.toString() !== "")) {
```
balcy, line 146: here the wrong id bookmarksFoldersViewItem (instead of bookmarksFoldersViewWideItem) was used, that's why it did not work on the tablet even if a homepage was set.

UniversalSuperBox (approved):

Issue #239, opened 2019-10-16
Pulldownmenus in webpages are incorrect


tux4us:
16.04 edge

When open a pulldownmenu from a webpage, the menu is to undersized to show all items. Menu don't adjust to numbers of items in menu.

balcy:
I think this issue could be moved to unity8
[edit: or mir ?]
see https://github.com/ubports/morph-browser/issues/54 for reference

lduboeuf:
or related to QtWebEngine ?

lduboeuf:
![66077182-bdc57900-e55f-11e9-8694-e5670bed69e5](https://user-images.githubusercontent.com/11663835/67748134-821cb280-fa2a-11e9-82cd-0016200a1be8.png)

lduboeuf:
Don't know if this can be fixed soon on Mir or QtWebEngine  ?
If not, can't we provide a js hook for `select` tags, and convert them to html elements stylable with css ?

Stefano0101:
Can confirm this is happening on all of my UT devices, Meizu Pro 5, Nexus 7 , ( Port Oneplus3t) on Egde channel, now after merge also on Devel.

giiba:
Switched to Devel channel to check out the new edge interface, but this issue is going to force me to switch back as it leaves the browser crippled for most sites that I encounter with drop boxes. The previous functionality was glitchy but I could at least proceed through online banking (for example).

giiba:
Switched to Devel channel to check out the new edge interface, but this issue is going to force me to switch back as it leaves the browser crippled for most sites that I encounter with drop boxes. The previous functionality was glitchy but I could at least proceed through online banking (for example).

On a OPO latest devel build.

giiba:
This seems to be a problem with the html form input 'select' tag, for a ready example see: https://www.w3schools.com/html/tryit.asp?filename=tryhtml_elem_select

Screenshot:
![screenshot20191103_100040082](https://user-images.githubusercontent.com/7121141/68089733-dfe13e00-fe20-11e9-8a01-3e4f8c03ffc7.png)



This doesn't seem to effect dropdowns done through javascript as those are working; see: https://www.w3schools.com/howto/tryit.asp?filename=tryhow_css_js_dropdown

giiba:
![screenshot20191103_094301874](https://user-images.githubusercontent.com/7121141/68089634-baa00000-fe1f-11e9-9c09-f6f1f1218a9c.png)

[...]

PR #273, opened 2019-12-19
Handle dropdowns (select html elements)

lduboeuf:
This is a workaround for html select boxes which does not display well currently (OTA-12).
The idea is to watch for select click events, pass them through a javascript prompt and use QML  to select options with a QML list overlay.


lduboeuf:
ref issue: https://github.com/ubports/morph-browser/issues/239

lduboeuf:
got some freeze on morph, need to know why

lduboeuf:
i had a Morph freeze today with playing with dropdowns. I'm not clear how to properly close a popup and make sure reject() is fired ( and received )  when dismissing the select popup

lduboeuf:
if someone can have a look to SelectOverrideDialog.qml

lduboeuf:
select

Example of rendering for youtube select filters. It works as a modal dialog

UniversalSuperBox:
Honestly I wouldn't say this is a workaround, but maybe the way we should have planned to do it from the beginning... So the select boxes look consistent with the rest of the system or other modals in Morph.

lduboeuf:
We could move the userScript to Morph.Web so that webapps can benefit later. and emit a custom signal. But don't know how to prevent signal onJavaScriptDialogRequested to be fired after

balcy:
maybe later QtWebChannel (https://doc.qt.io/qt-5/qtwebchannel-javascript.html) can be used instead of the java script dialog to trigger the method. Right now the qml module of it is not part of the read only images though.

lduboeuf:
yes was thinking of that too

cibersheep:
A W E S O M E !

lduboeuf:
Todo: handle pre selected option

lduboeuf (commented):
addEventListener should be on document.body, no need to be higher

lduboeuf:
tried to work with ListView instead of a Repeater, it is faster, and can easily point to the pre selected option, but still have height issue for the popup as it does not grow automaticaly with ListView

src/app/SelectOverrideDialog.qml
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2013-2016 Canonical Ltd.
+ * Copyright 2016-2020 Ubports
```
balcy: @UniversalSuperBox can "Ubports" be used as copyright owner by now ?

UniversalSuperBox: "UBports Foundation" is more correct, but valid

src/app/PromptDialog.qml
```diff
@@ -30,9 +30,6 @@ ModalDialog {
     signal accept(string text)
     signal reject()
     
-    onAccept: hide()
-    onReject: hide()
-
```
balcy: those should not be removed IMO.
Right now it works without them, because within one function only one dialog is used, and it disappears when the respective function that created the dialog object is finished (the object goes out of scope).
I've tested it with a prompt dialog and another dialog in the same function

                var promptDialog = PopupUtils.open(Qt.resolvedUrl("PromptDialog.qml"), this);
                var alertDialog = PopupUtils.open(Qt.resolvedUrl("AlertDialog.qml", this))

then you can not close the prompt dialog any more by pressing OK or Cancel. Also if there is sth, that takes longer to process in the same scope, the dialog would stay open during that time.
All other dialogs do have the respective `hide()` or `PopupUtils.close(dialog)` methods, too.

balcy (changes requested):

balcy:
the feature itself did work great for me on the phone so far.

src/app/PromptDialog.qml
lduboeuf: Yes it was what i've used before ( hide() on accept signal in the dialog), but for some reason i had webpage freeze (even tab freeze) when doing that on big dropdowns. ( see https://lduboeuf.github.io/select.html for tests) .

lduboeuf: i'm going to test without hide()

lduboeuf: well no luck

balcy: ok, but for the PromptDialog the old way can be kept.

lduboeuf: yes seems not a problem for other cases

balcy (dismissed):
ok thanks ! Since the prompt dialog is used at various places (e.g. adding new urls for domain permissions), I wanted to keep the used behavior here.

gbdomubpkm:
The fix isn't on the dev channel yet, is it?
Here's a good page for one last test I couldn't perform: https://www.prix-portables.fr/#-

lduboeuf:
no but you can safely test it by installing this PR : `sudo ubports-qa install PR-273` , its seems to work well on this site

lduboeuf:
here is my tests ( but don't cover all cases ) : https://lduboeuf.github.io/select.html

gbdomubpkm:
I can´t install the PR on my OPO 64 Go.
I get this : 
PR-273

lduboeuf:
> I can´t install the PR on my OPO 64 Go.

just do that before :
`sudo mount -t tmpfs tmpfs /var/cache/apt/archives`

or if you just want to test morph without replacing the current one, just clone the PR and run clickable ( you will have a second browser )

gbdomubpkm:
Great! Nice! I tried on different sites and works very well in morph. A big annoying bug solved. 👍
Thank you!

However, do you know that the correction seems inoperative in webapps: for example, having tested in Mateo Salta's youtube webapp, the subtitle option can't be chosen for a video yet.

Do I have to remove PR-273 or will a future morph update in dev write over it?

lduboeuf:
Yes it does not work work webapps yet. It would need some extra work and don't find how to do this. 
Hope that this current fix will land soon on devel  but that is not my decision :)

gbdomubpkm:
I was partially wrong  : Good news ! I created a youtube webapp with Tim Suberkrub's Webber app and your system works fine ! (at least with webber).

lduboeuf:
Oh awesome, but that means that i don't understood what part of Morph.Web webapps embed :)

balcy:
I think "simple" webapps (calling webapp-container with arguments in their .desktop file) do work, but apps with custom QML using Morph.Web as library don't.

lduboeuf:
ok i see, thx. 

I get stuck at moving things to Morph.Web, by handling onJavascriptDialogRequested there, seems i can't forward signal ( it leads to a Morph Crash ), was thinking that `accepted` to false will do the trick also ( if select then open popup else, let the container do the things ) but no, it tries to load default popups but not installed on UT
 Another solution would be to write a full js css solution, but will be hard to theme no ?

[EDIT]: humm i've probably made a mistake somewhere, i'm pretty close to get it running by moving parts to Morph.Web but of course need to add a property and handle it on WebViewImpl I keep you updated

lduboeuf:
so does not work yet :)

lduboeuf (commented):
not sure it is a good thing to make it a Singleton as i guess will slow the initial load of Morph Browser ?

lduboeuf:
The lasts commits moves "select-override" management to Morph.Web so that apps can benefit from it.
For a backward compatibility, Morph.Web don't do nothing new by default. this feature is enabled by setting `enableSelectOverride` to true, the drawback of this  is that developer must check if the DialogRequest has not already been handled  ( https://github.com/ubports/morph-browser/pull/273/files#diff-e2100084e31ab20d6ea7cd28e7875dd6R92 )

`sudo mount -t tmpfs tmpfs /var/cache/apt/archives`
`sudo ubports-qa install PR-273`


for a reminder: past workable solution (but only for Morph Browser and webapp-container ) here: https://github.com/ubports/morph-browser/pull/273/commits/e6f7e1b010ea44d974bf814e1b7ad5c670511143

lutin11:
thanks, greet work !

Flaburgan:
hi @lduboeuf and thank you for your work, this patch is very needed as not being able to use `<select>` elements is a huge regression compared to OTA-11. 

I wanted to try it on my phone but `apt` says it will downgrade the package. Is that expected? Or do you need to rebase your PR on the current code?

lduboeuf:
humm, this PR seems to have no conflict. You can try it safely

lduboeuf:
@balcy  @UniversalSuperBox : hello reviewers, should this PR be merged  one day or any other viable solution to come ?

gbdomubpkm:
@lduboeuf My understanding is that it is probably not yet possible to examine dev porting. There must be some other points to be settled beforehand on which the behaviour of your magnificent mechanism could depend. Non-specialist opinion maybe i'm wrong ! 😉

balcy:
I'd let Dalton decide when it should be merged

UniversalSuperBox:
I'm sorry, I'm still out for events. If you feel this is ready for merge after your testing, @balcy, please do so. :)

lduboeuf:
@balcy just let me clean a bit the changes tomorrow to make the diff smaller 
One question: Should this feature be enabled by default or let the Morph.Web user decide ? ( currently this feature is not enabled by default, user need to enable it explicitely )

lduboeuf:
ok done

Flaburgan:
>ok done

Looks like something didn't work, you still have 23 commits :p

You can squash them by doing `git rebase -i xenial` then replace `pick` by `r` in front of the first commit and `f` for the others. Then you will be able to reword your first commit with a message like `Allow morph to handle <select> elements with a custom dropdown`. Don't hesitate to ping me on telegram if you need help with git.

lduboeuf:
@Flaburgan thanks for the hint. I let the reviewer squash my commits when merging as i think its always done for all project here :)

Flaburgan:
Oh okay, it's usually done by the developers in all the other projects I'm contributing to.

balcy:
yes opt in is fine IMO (not enabled by default in Morph.Web)

lduboeuf:
ok, so should be fine :)

src/Morph/Web/MorphSelectOverrideDialog.qml
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright 2013-2016 Canonical Ltd.
+ * Copyright 2016-2020 Ubports Fondation
```
UniversalSuperBox: "UBports Foundation"

src/Morph/Web/select_overrides.js
```diff
@@ -0,0 +1,37 @@
+/**Workaround for https://github.com/ubports/morph-browser/issues/239 ( OTA-12 ) Until it is fixed in low level.
```
UniversalSuperBox: No need for two asterisks in comments here, two asterisks marks the comment as a documentation comment. Documentation is not built from javascript files.

src/app/WebViewImpl.qml
```diff
@@ -89,6 +89,8 @@ WebView {
 
     onJavaScriptDialogRequested: function(request) {
 
+        if (isASelectRequest(request)) return; //this is a select box , Morp.Web handled it already
```
UniversalSuperBox: Morph.Web

src/Morph/Web/MorphWebView.qml
```diff
@@ -48,6 +48,42 @@ WebEngineView {
         certificateError = null
     }
 
+    /**
+       html select override
+       set enableSelectOverride to true to make Morph.Web handle select
+       note that as it uses javascript prompt,
+       make sure that onJavaScriptDialogRequested signal handler don't overplay prompt dialog by checking the isASelectRequest(request)
+      **/
```
UniversalSuperBox: Note the format of other documentation comments in this file:

* Each line begins with an asterisk
* There is a distinct character limit
* The comment ends with a single asterisk and a slash

Please replicate https://github.com/ubports/morph-browser/pull/273/files#diff-f88008971a9cf96598fe24671fe22d53R87, for example.

src/Morph/Web/MorphSelectOverrideDialog.qml
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    //make sure reject is fired when closing the popup
+    Connections {
+        target: __eventGrabber
+        onPressed: selectOverlay.reject()
+    }
+
+
+
```
UniversalSuperBox: Any reason for such a large space at the end of this file?

UniversalSuperBox (changes requested):

src/Morph/Web/MorphSelectOverrideDialog.qml
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2013-2016 Canonical Ltd.
+ * Copyright 2016-2020 UBports Fondation
```
UniversalSuperBox: Fo**u**ndation

If this code was not copied from another file with a similar copyright, there is no need to have the 2013-2020 copyright chain. Instead, just `2020 UBports Foundation` is enough.

UniversalSuperBox (changes requested):
Fixing this copyright header should be all. :D

UniversalSuperBox (approved):
Excellent work. Thank you!

PR #274, opened 2019-12-20
Xenial   startprivate

### LINKED ISSUE #145: Feature: Allow to start browser in private mode
It would be nice to be able to start browser in private mode.
I prefer such way of using browsers, for instance, cookies are not stored between sessions.

balcy:
fixes https://github.com/ubports/morph-browser/issues/145

### DIFF
--- a/src/app/webbrowser/morph-browser.qml
+++ b/src/app/webbrowser/morph-browser.qml
@@ -30,11 +30,11 @@ QtObject {
 
     function init(urls, newSession, incognito) {
         i18n.domain = "morph-browser"
-        if (!newSession && settings.restoreSession && !incognito) {
+        if (!newSession && settings.restoreSession && ! (incognito || settings.incognitoOnStart)) {
             session.restore();
         }
         if (allWindows.length == 0) {
-            windowFactory.createObject(null, {"incognito": incognito}).show();
+            windowFactory.createObject(null, {"incognito": (incognito || settings.incognitoOnStart)}).show();
         }
         var window = allWindows[allWindows.length - 1];
         for (var i in urls) {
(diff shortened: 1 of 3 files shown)


UniversalSuperBox:
This Jenkinsfile change does not work -- the server it's building on is not labeled armhf

balcy:
ok undone the change

src/app/webbrowser/morph-browser.qml
```diff
@@ -30,19 +30,19 @@ QtObject {
...
+            windowFactory.createObject(null, {"incognito": (incognito || settings.incognitoOnStart)}).show();
         }
         var window = allWindows[allWindows.length - 1];
         for (var i in urls) {
             window.addTab(urls[i]).load();
             window.tabsModel.currentIndex = window.tabsModel.count - 1;
         }
         if (window.tabsModel.count === 0) {
-            window.addTab(incognito ? "" : settings.homepage).load();
+            window.addTab(settings.homepage).load();
```
UniversalSuperBox, line 45: Should the homepage be loaded even if the user wants to be in private mode? I'm not sure.

balcy:
I was not sure, too, and asked it in the issue.
Assuming we treat morph started with the --incognito flag the same as morph started with the new "startprivate" flag set, not showing the homepage would mean using the new option prevents the possibility of a custom homepage for those users. (only Open new window -> Homepage would bring it up)
If a user sets a homepage they likely trust it IMO.
What about the following: We revert the line to the old state (--incognito flag is without homepage) but the "startprivate" mode is with homepage.
When the user activates the startprivate checkbox and has a homepage set, we ask wether to remove (set to "") or keep the homepage.
Alternatively we show the homepage neither with --incognito nor in startprivate mode.

UniversalSuperBox:
It's a mixed bag on desktop, too. If you set Firefox to "Always use private browsing", it allows you to set a homepage after it restarts and keeps that homepage. If you start Chromium in incognito mode via the command line option (there's no other way), it always opens to the warning screen.

So, either seems fine.

UniversalSuperBox:
I believe the current behavior is acceptable, if you have no further changes to make.

balcy:
ok no changes left I think!

UniversalSuperBox (approved):

PR #281, opened 2020-01-04
Fix the weird shadow behavior on mobile

### LINKED ISSUE #11: tab shadow is longer than the tab
open 2-3 tabs, swipe from the buttom to see all the tabs. the shadows are longer than the tabs

next

bhdouglass:
Fixes the tab's shadow to be only behind the tab, and not flying off to the side. I'm not sure why the BorderImage wasn't working as expected, so I just dropped that and made it a regular image. This does lose the shadow on the top of the tab preview. I replaced that with just a border line. If that doesn't work I can see about bringing back the shadow.

Ping @cibersheep for design review.

Fixes #11 

Before:

screenshot20200104_141736342

After:

screenshot20200104_141618074


cibersheep:
I can't believe I didn't see that bug being there any more...

Looks good to me. Have you tested on the tablet?

bhdouglass:
Yep! Here is a screenshot: 
screenshot20200106_005118647

balcy (approved):

cibersheep:
Awesome!

PR #284, opened 2020-01-25
Relax the BrowserWindow minimumWidth

mariogrip:
Relax the BrowserWindow minimumWidth as some devices have a width
smaller than the minimum causing the app to overflow.


balcy (approved):
seems to be OK on BQ Aq. 4.5

PR #294, opened 2020-02-14
Add dim background color to select element popup

giiba:
First, thanks to @lduboeuf for this implementation! This is a great looking way to handle these menus, much easier to interact with than the tiny box that shows up in Demo Browser.

However I am curious why the dimming of the background was disabled? On a white page it makes the select dialogue look very undefined... I think at a minimum there should be a border to define the popup, or just letting the site be dimmed which I personally like.

My suggestion:
screenshot20200214_071231219

As exists now:
screenshot20200214_074419061

I know there is a slight border, but it doesn't feel like enough. Just a suggestion.

### DIFF
--- a/src/Morph/Web/MorphSelectOverrideDialog.qml
+++ b/src/Morph/Web/MorphSelectOverrideDialog.qml
@@ -27,7 +27,7 @@ Popups.Dialog {
     objectName: "selectOverrideDialog"
     modal: true
 
-    __dimBackground: false //avoid default opaque background
+    //__dimBackground: false //avoid default opaque background
     grabDismissAreaEvents: false //allow this component to handle the click on the background
 
     property string options: ""


lduboeuf:
We should ask UX team if ok with default opaque bg color for dropdown, this was only my personnal opinion. cc @cibersheep ?

Or we need to change Popup component in ui toolkit a bit to allow the background customization.

cibersheep:
Interesting question. The default opacity depends on the size of the screen. On mobile is too dark to see what we are selecting in the pop up but it's the standard behaviour (the toolkit expects it underneath that's why the border is so subtle).

Eather go with standard or as lduboeuf says: a less dark layer

lduboeuf:
@cibersheep , see PR https://github.com/ubports/ubuntu-ui-toolkit/pull/68 for allowing customization of the background color

cibersheep:
I don't think we should let change the color of the background dim but the opacity.

lduboeuf:
@cibersheep why not ? If any apps ( others than the core ones ) want to play with a different style, as we can both set color and opacity with the color property

cibersheep:
I see your point. I might have been too afraid of people :)

lduboeuf:
@gliba, could you please edit the PR title to be more precise, e.g "add dim background color to dropdowns"  ?

kugiigi:
Old PR but can't we use Popup/Dialog from qqc2? I think they are more flexible and the dimming isn't as opaque as in UITK. Actually, this is more of a general question because I have things in mind where I intend to use qqc2 instead of UITK.

dobey:
Is this good too merge @cibersheep ?

cibersheep:
We have changed the SuruDark Dialog colors recently https://github.com/ubports/ubuntu-ui-toolkit/pull/98
I think the dim background could behave like the one under the Dialog

dobey:
> We have changed the SuruDark Dialog colors recently [ubports/ubuntu-ui-toolkit#98](https://github.com/ubports/ubuntu-ui-toolkit/pull/98)
> I think the dim background could behave like the one under the Dialog

Can you please update the branch in this PR according to this comment @giiba ? Thanks.

PR #307, opened 2020-03-25
Xenial   pdfextension

### LINKED ISSUE #148: open downloaded items in browser
Currently the downloaded items (downloads page) have the following user interactions:

- normal click / tap : "Open with" menu (content hub) opens, file can be opened with another app
- long press: enable "select mode", which allows to delete multiple downloads

Especially for webpages downloaded as HTML, but for text and multimedia files as well, it would be useful to open them directly in the browser after downloading, or at some point later.
Moreover it could be useful to copy the path or the *file://...* url of the downloaded file to the clipboard.

So the question would be how could this be integrated to the downloads menu.

Possible solution (1)
- normal click / tap : "Open with" menu (content hub) opens, file can be opened with another app
- long press: show dialog:
  - open with other app
  - open in browser
  - copy path
  - copy file url
  - edit items

Possible solution (2)
- normal click / tap : show dialog:
  - open with other app
  - open in browser
  - copy path
  - copy file url
- long press: enable "select mode", which allows to delete multiple downloads

or something else ?

### LINKED ISSUE #254: Downloads for webapps: simplify code for QtWebEngine 1.9 (Qt 5.13)
since we currently use an older version of QtWebEngine the following property is not yet available:
https://doc.qt.io/qt-5/qml-qtwebengine-webengineprofile.html#downloadPath-prop

So QtWebEngine uses the default path for downloads (/home/phablet/Downloads), which the webapps do not have write permissions to.

What is possible with the current version of QtWebengine (Qt 5.11) is setting the download path for each file to be downloaded.
https://doc.qt.io/qt-5/qml-qtwebengine-webenginedownloaditem.html#path-prop
Therefore we have to check if the file does exist ourselves and select another name if it exists (e.g. file.txt, file(1).txt, ...)

This issue is a reminder to remove that again as soon as we move to QtWebEngine 1.9 (Qt 5.13) and just set the download folder in the beginning.

As a reminder this issue is mentioned in
src/app/webcontainer/webapp-container.qml
src/app/webcontainer/WebApp.qml

### LINKED ISSUE #306: "canGoBack" change does not propagate when changing the url (e.g. via address bar)
- open morph with the homepage (e.g. start.duckduckgo.com)
- write another url (e.g. ubports.com) in the address bar

expected result: the back button is enabled
actual result: the back button is not enabled

I don't know why the change of the flag is not properly propagated:
In fact if you manually evaluate webview.canGoBack after navigating to the page, it is "true", but the property is not automatically updated.

balcy:
fixes https://github.com/ubports/morph-browser/issues/306
fixes https://github.com/ubports/morph-browser/issues/254
fixes https://github.com/ubports/morph-browser/issues/148

- adaptions for QtWebEngine 5.14
- preview PDFs
- open downloads in browser
- add url parameter to downloads
- improve PDF creation (delete temporary PDFs)


balcy:
https://github.com/ubports/qtwebengine-opensource-src-packaging/pull/30 has to be merged before this morph version can be used.

balcy:
there have been problems with merged .po files, I'll open a new (cleaner) PR

ghost:
Tested with FP2 : everything works, thanks a lot !! :fireworks:

jezek:
Device: FP2
Image: 2020-W36

Tested:
- Browser can go back (#306)
- Webapp (google translate) has download folder set to ```~/.local/share/$appname```. Is this sufficient as test for (#254)?
- Downloaded apps actions (#148)
  - Normal click opens hub. On some files (jpg, mhtml, txt, pdf) on the right end of "Open with", there is an icon (folder with sheet of paper upfront), which opens item in browser (pdf works). On some files, which contain downloaded url (mhtml, txt) here is one more icon (rounded square with arrow coming from middle to right top corner and out of the square), which opens the url. I've spotted small bug in opened content hub. If clicking on the "Open with" text, the click event is catched by the download list underneath and changes the content hub for the first item in downloads (it happens to be under the "Open with" text)
  - Long click opens select. Downloads can be selected and deleted as expexted.
  - Swiping download right allows to delete.
  - Swiping left doesn't show "copy path" or "copy url" as suggested by @cibersheep in https://github.com/ubports/morph-browser/issues/148#issuecomment-466687023

Anything else to test?

PR #323, opened 2020-06-11
Detect QtWebEngine's Chromium version from its default UA

peat-psuwit:
As long as we don't overwrite it, we can get the default user agent from
WebEngineContext and parse Chrome/Chromium version out of it. Do that on
Component.onCompleted before setting our user-agent.

Goes along well with https://github.com/ubports/qtwebengine-opensource-src-packaging/pull/30

### DIFF
--- a/src/Morph/Web/MorphWebContext.qml
+++ b/src/Morph/Web/MorphWebContext.qml
@@ -28,12 +28,12 @@ WebEngineProfile {
     property alias incognito: oxideContext.offTheRecord
     readonly property string defaultUserAgent: __ua.defaultUA
 
+    property string chromiumVersion
+
     dataPath: dataLocation
 
     cachePath: cacheLocation
     maxCacheSizeHint: cacheSizeHint
-
-    userAgent: defaultUserAgent
 
     persistentCookiesPolicy: WebEngineProfile.ForcePersistentCookies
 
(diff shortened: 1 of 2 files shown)


balcy (dismissed):
looks good to me

balcy:
while I like the idea of dynamically determining the chromium version, it will change only once in a while, and thus can be provided as a constant. With the git command it can be read: https://wiki.qt.io/QtWebEngine/ChromiumVersions

peat-psuwit:
- Isn't the git command supposed to be run on the QtWebengine's tree?
- Regarding the "does not include 'mobile'", could you please elaborate?

balcy:
I've installed the version of the PR on the phone, and then "https://duckduckgo.com/?q=user+agent&ia=answer" did not include "Mobile" (it behaved like being set to desktop mode). That might be a timing problem because UserAgent02.qml maybe is not ready when sending the request to the server.
At least on my phone I got the desktop version. Therefore I've removed it from the branch I had merged it for now.
I think it would be nice to set it dynamically, but since it does not change very often, we could keep it static.
Your way is much better for testing one morph with different QtWebEngine versions, so maybe we find a dynamic way.
You're right the git command would only help at build time, and only for one version.

balcy:
I've slightly modified it (used a temporary WebEngineProfile object):
https://github.com/ubports/morph-browser/commit/711b2fddec1ba0d406f7b4c13416a6079a8bf168

balcy:
it was integrated (slightly modified) in https://github.com/ubports/morph-browser/pull/328

PR #327, opened 2020-06-28
Xenial   qtwebengine514

balcy:
redone https://github.com/ubports/morph-browser/pull/307


balcy:
@UniversalSuperBox I think the user experience is better with high dpi scaling enabled, because some pages don't really adapt well to the zoom.

PR #331, opened 2020-07-08
SuruDark fixes

kugiigi:
* Some colors were changed even in Ambiance
* The highlight color changed to Orange because Purple doesn't play well in dark mode

### DIFF
--- a/src/app/webbrowser/BookmarksView.qml
+++ b/src/app/webbrowser/BookmarksView.qml
@@ -73,7 +73,7 @@ Common.BrowserPage {
             }
 
             activeFocusOnPress: false
-            strokeColor: theme.palette.selected.base
+            color: theme.palette.normal.base
             text: i18n.tr("Done")
 
             onClicked: bookmarksView.back()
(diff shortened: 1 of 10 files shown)


mateosalta:
perhaps this should be the blue color, 'focus' ? this will have a bright blue in suru dark , and regular blue in ambiance.

kugiigi:
It's actually the first one I considered and tried but somehow I feel like blue is already used too much in Morph. I don't know just my opinion. Perhaps someone from the design team? 😄

mateosalta:
blue does make sense for url's, and we are  on an orange purge lately ;) perhaps we could use a different grey? 
@cibersheep what do you think?

cibersheep:
- Do we have some screenshots?
- About orange for highlight: that's the old highlight color. Purple is never used before. For selected and focus you have the theme palette colors for that :) now are indeed blue 
- Note to myself: I need to fix the stroke color buttons now! :(

kugiigi:
Here are some screenshots:

**Current implementation**
screenshot20200722_234807646

screenshot20200722_234519391

**Using focus palette**
screenshot20200722_235404447
screenshot20200722_235523598

**Using focus palette and NOT bold**
screenshot20200722_235809038
screenshot20200723_000010803

cibersheep:
I see. Thank you.
I would definitevely go for focus + bold

kugiigi:
I've updated the highlight palette to use `theme.palette.normal.focus`

dobey (approved):
Looks OK.

Issue #165, opened 2019-03-03
Scroll up navigation bar


balcy:
The webbrowser-app did have the feature, that when scrolling down a website, the navigation bar scrolled out of view upwards.
Therefore more space for the content was available.
When scrolling up, the header bar scrolled back into view.
We could try to reimplement that at some point.

cibersheep:
In theory, is adding the property `flickable: theIDofTheFlickableElement` here: 
https://github.com/ubports/morph-browser/blob/xenial/src/app/webbrowser/BrowserPage.qml#L43

krille-chan:
The QtWebEngine has a property "scrollPosition" where we can watch at scrollPosition.y to determine if the user has scrolled down to hide the navigation bar: https://doc.qt.io/qt-5/qml-qtwebengine-webengineview.html#scrollPosition-prop

josele13:
that would help a lot to see the full screen, but that would be necessary also in the webapps, because by default there is a space that is not seen of the web page in its inferior zone, the web page is not complete, you can try for example with My Ut forum

cibersheep:
Could you add a screenshot @josele13 ?

josele13:
Yes of course...
One of the photos is Morph, the other is the webapp, both down the page to the maximum possible. The webapp does not open the whole page, the space occupied by the top bar,

greetings...
![screenshot20200611_234416237](https://user-images.githubusercontent.com/52839437/84443758-91124700-ac40-11ea-95d6-08918a54f453.png)
![screenshot20200612_000449229](https://user-images.githubusercontent.com/52839437/84443770-97082800-ac40-11ea-8547-05bd9ee0feee.png)

![screenshot20200611_235145176](https://user-images.githubusercontent.com/52839437/84443868-ccad1100-ac40-11ea-99cc-9a553fdc80a8.png)
![screenshot20200612_000005237](https://user-images.githubusercontent.com/52839437/84443872-ce76d480-ac40-11ea-8f1f-d6e56f87b93f.png)

mauricioduarte01:
> The webbrowser-app did have the feature, that when scrolling down a website, the navigation bar scrolled out of view upwards.
> Therefore more space for the content was available.
> When scrolling up, the header bar scrolled back into view.
> We could try to reimplement that at some point.

Hello,
Balcy, any news regarding this feature?
It would be really important to have this implemented in terms of usability.
Thank you in advance for your hard work

balcy:
[...]

PR #332, opened 2020-07-08
Auto hide the bottom edge hint

kugiigi:
Behavior of the bottom edge hint
- Hides after 2 seconds of not scrolling
- Always shown while scrolling
- Initially won't auto-hide until the webview is scrolled the first time (same behavior after switching to a tab)

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -986,6 +986,11 @@ Common.BrowserView {
         }
         visible: bottomEdgeHandle.enabled && !internal.hasMouse
         opacity: recentView.visible ? 0 : 1
+
+        function show() {
+            opacity = Qt.binding(function(){return recentView.visible ? 0 : 1})
+        }
+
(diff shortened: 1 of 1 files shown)


balcy:
I think the bottom edge hint is hidden by the ScrollTracker (still code of webbrowser app right now that had not been adapted to morph yet):
With that commit both the chrome and bottom edge hint should be hidden, as I remember. (They come back if the bottom or top of the page is reached)
https://github.com/ubports/morph-browser/commit/57e09afce5f45772ce6afd69ce399d8f6c99df5e

(see https://github.com/ubports/morph-browser/issues/165)

kugiigi:
Really? I'm not aware of this or at least I don't remember. Anyway, if it does indeed then that's fine :smiley:

UniversalSuperBox:
@balcy, is this superseded by https://github.com/ubports/morph-browser/pull/333?

UniversalSuperBox:
I believe this is superseded by #333.

PR #333, opened 2020-07-09
adapt ChromeStateTracker / ScrollTracker for QtWebEngine

### LINKED ISSUE #165: Scroll up navigation bar

balcy:
fixes https://github.com/ubports/morph-browser/issues/165


dobey:
What's the status on this?

balcy:
that's still draft let's wait with it a bit (at least not ready for OTA14)
It can happen that the menu keeps showing/hiding at certain scroll positions.

kugiigi:
Perhaps we can use a similar approach as this?
https://gitlab.com/mateosalta/qtwebbrowser_click/-/blob/dev/src/qml/PageView.qml

It's not tracking scrolling per se though but it works pretty well in my experience.

mateosalta:
> Perhaps we can use a similar approach as this?
> https://gitlab.com/mateosalta/qtwebbrowser_click/-/blob/dev/src/qml/PageView.qml
> 
> It's not tracking scrolling per se though but it works pretty well in my experience.

only problem i saw with that one is the page extends off the screen until you scroll

kugiigi:
What are the worst case with that issue? Like when a page doesn't scroll vertically and takes up the whole viewport? I think in most cases, it'll be fine since you naturally scroll anyway to see the bottom part.

balcy:
ok we might give it a try, I'm still not sure if it usable like that, any improvements / suggestions welcome.
We really could mimic how it is done in "WebEngineDemo", it is quite good to use.

Another question: should we make it an option to automatically hide the Chrome when scrolling down, or activate it for all users ?
It used to be like that for the old webbrowser-app (without an option).

PR #334, opened 2020-07-09
Create ActionList after the place that refers it

peat-psuwit:
This is to workaround QObject children destruction order which destructs
in creation order. Otherwise, the ActionList and its children would be
destroyed before e.g. the Repeater, which will cause SIGSEGV at exit
when the Repeater is subsequently destroyed.

I don't understand why it's done this way; I would expect objects
declaratively created to be destroyed in the reverse order, in the same
way objects created on the stack in C++ would.

Also contains a code cleanup in a separate commit.


balcy:
ah so that is where the segfault on close came from, interesting!

UniversalSuperBox (approved):

jezek:
There is a call for testing, how exactly test this?

peat-psuwit:
If the Morph browser works correctly for you, then it should be fine. Try bringing the selection control (copy/select all etc.) up.

jezek:
Device: FP2
Image: 2020-W37

Morph browser seems to work correctly. Brief tests performing copy/paste actions were working without any problems.

Issue #123, opened 2018-12-02
Browser downloads should be integrated with indicator-tranfer


cibersheep:
**OS:** Xenial RC
**Steps:** Tap on a click to download a file (for example, last issue on Full Circle Magazine, http://dl.fullcirclemagazine.org/issue139_en.epub )
**Behaviour:** Morph starts downloading the file showing downloads list page. Pulling files indicator shows nothing.
**Expected:**
1) To be asked if user wants to download the file (specially if the same tab is already open and Morph is restarted)
2) Download being handled by DownloadManager so it shows in the file indicator

balcy:
yes both would be good features.
For 1) I thought it would be easy to implement, but for some reason the request did not like being delayed that way. (https://doc.qt.io/qt-5/qwebengineprofile.html#downloadRequested)
2) is unfortunately not possible with the currently used implementation (https://doc.qt.io/qt-5/qml-qtwebengine-webenginedownloaditem.html#properties).
It does not provide the url property, which might be a Qt bug (https://forum.qt.io/topic/94571/webenginedownloaditem-url-property)
But even with this property, by using the ubuntu download manager, we can no longer use the builtin mechanism / download queue provided by QtWebEngine itself. (see https://github.com/ubports/morph-browser/pull/32), so it will take additional work (e.g. getting the current session for the download)
But you are right the currently it is for example not possible to have background downloads with morph, and being asked for the same download again is not ideal of course.
Hopefully at least for 1) we find a solution, e.g. accepting the download, pausing it, and then ask the user.

cibersheep:
Oh, I see. I didn't expect it would be so complicated :)

PR #336, opened 2020-07-11
Changes in Downloads UI/UX

### LINKED ISSUE #193: Annoying downloader
The downloader not only opens when you start a download, but also when it finishes downloading.
This interrupts your browsing experience.

It would be much less annoying if the browser just showed a litte "Download of ... started" on the open bottom of the screen.
A finished download could be indicated, either by a similar message on the bottom of the screen, or by just using the notifications at the top.

### LINKED ISSUE #330: Deleting a download does not delete the file
## Steps to Reproduce
1. Download a file
2. Wait until it completes (Getting an error is also fine)
3. Delete the completed/failed download

## Expected Results
The download is delete from the list and the corresponding file is deleted

## Actual Results
The download is delete from the list and but the corresponding file is retained.

## Additional information
I'm not sure if this is intended but if it is, perhaps an option to delete both is an option.
I'm also actually checking the code right now so I can do a fix myself :)

kugiigi:
- Fixes #330
- Fixes #193
- ~Fixed issue in webapps with title bar overflows content in the bottom~ (fixed in https://github.com/ubports/morph-browser/pull/351)
- Implemented recent downloads dialog in bot the browser and webapp. This is displayed when a new download started instead of the downloads page.
- A downloads button is added in the navigation bar which wiggles and turns blue when a download completes. In webapps without the title bar, the recent downloads dialog is displayed instead
- Only display the file names in the Downloads page instead of full path.
- Added info such as file size and download speed in the Downloads page
- Put the content picker when opening downloads into a popup for a more "convergent" design.


mateosalta:
sounds cool, can you attach some pictures :)

kugiigi:
I was supposed to attach some but apparently I completely forgot 😅
Here's a couple.
screenshot20200721_025816857
screenshot20200721_025825117

balcy:
see PR https://github.com/kugiigi/morph-browser/pull/1 that rebased the branch to recent merged xenial (includes xenial_-_pdfextension) and moved "open file / link in browser" actions to ContentExportDialog.
This should after merging have no conflicts with the current xenial branch.

balcy:
oh, almost forgot: thank you for the huge redesign effort !

src/app/webcontainer/WebApp.qml
```diff
@@ -212,6 +275,7 @@ Common.BrowserView {
                 left: parent.left
                 right: parent.right
                 top: chromeLoader.bottom
+                bottom: parent.bottom
```
balcy: this has been resolved by https://github.com/ubports/morph-browser/pull/351, binding it to parent.bottom would hide the part of the webview below the on screen keyboard if it is open. (in https://github.com/kugiigi/morph-browser/pull/1 it is therefore removed)

balcy:
... I forgot to connect some parts to the DownloadsDialog, I'll add them in a bit.

balcy:
A question: should the downloads icon in the chrome be hidden when the current downloads are cleared by the user ?

kugiigi:
> A question: should the downloads icon in the chrome be hidden when the current downloads are cleared by the user ?

No, once a download is started, the download icon is displayed until you restart the app. I believe this is the same behavior in Firefox.

balcy (changes requested):
https://github.com/kugiigi/morph-browser/pull/2 should be merged first

balcy (dismissed):

cibersheep:
I think we should use the transfers indicator :)
- is perfect for displaying downloads in an unobstructive way (like when you download a podcast with Podbird)
- Has already pause / resume implemented
- Doesn't block the UI

https://github.com/ubports/indicator-transfer

kugiigi:
> I think we should use the transfers indicator :)
> 
> * is perfect for displaying downloads in an unobstructive way (like when you download a podcast with Podbird)
> * Has already pause / resume implemented
> * Doesn't block the UI
> 
> https://github.com/ubports/indicator-transfer

That's what I originally wanted but it seems that it needs more work and there were bugs that makes it complicated to do.

cibersheep:
I understand. Then we need to address those bugs :) 
I mean, I appreciate your work.

Do we know which bugs we need to fix?

kugiigi:
You have some discussions in #123  😊
The `url` property seems to be working now with the updated QtWebEngine so we're a bit closer to make it possible.
But I think @balcy has more knowledge about this.

cibersheep:
Thanks <3
It looks like my recurrent subject :)

mateosalta:
> > A question: should the downloads icon in the chrome be hidden when the current downloads are cleared by the user ?
> 
> No, once a download is started, the download icon is displayed until you restart the app. I believe this is the same behavior in Firefox.

for a desktop browser maybe. I think for morph we should hide it as much as possible
 - show during and after download
- I agree that it should hide if you clear recent downloads and have no current downloading

balcy:
I think it can be merged, only minor thing (comes from my commit) is maybe the background color of the 'open in browser' label and icons, but should be OK for now.

balcy (dismissed):

balcy:
if I merge this, does it have to be linked in the project (OTA14 QA)?

UniversalSuperBox:
The linked issues should be placed on the board, not the PR itself.

However, we are in OTA-14 release freeze. Now is not a good time to merge. After the OTA-14 release, we're opening back up for OTA-15. This PR can be part of that.

src/app/ContentExportDialog.qml
```diff
@@ -0,0 +1,120 @@
...
+            id: openInBrowserRow
+            anchors.top: dialogItem.bottom
+            anchors.right: parent.right
+            spacing: units.gu(1)
+            padding: units.gu(1)
+
+        Label {
+            id: labelOpenInBrowser
+            visible: openFileInBrowser.visible || openLinkInBrowser.visible
+            text: i18n.tr("open in browser")
```
UniversalSuperBox: Any reason why there is no capitalization in "open in browser"?

Also, since you're adding new strings, we'll need to rebuilt the translation files and add them to this PR.

src/app/DownloadDelegate.qml
```diff
@@ -132,71 +167,59 @@ ListItem {
...
-                width: cancelButton.width
-                onClicked: {
-                    if (download) {
-                        download.resume()
+                    
+                    Label {
+                        horizontalAlignment: Text.AlignHCenter
+                        textSize: Label.Small
+                        // TRANSLATORS: %1 is the percentage of the download completed so far
+                        text: "(" + i18n.tr("%1/s").arg(FileUtils.formatBytes(downloadDelegate.speed)) + ")"
```
UniversalSuperBox, line 205: It doesn't seem like there is any translatable text here. Why is it a translatable string?

src/app/DownloadsDialog.qml
```diff
@@ -0,0 +1,190 @@
...
+
+            property int selectedIndex: -1
+            
+            delegate: SimpleDownloadDelegate {
+                download: modelData
+                title.text: FileUtils.getFilename(modelData.path)
+                subtitle.text: if (cancelled) {
+                            i18n.tr("Cancelled")
+                        } else {
+                            (error ? modelData.interruptReasonString : (incomplete ? i18n.tr("%1%").arg(progress) : i18n.tr("Completed")))
```
UniversalSuperBox, line 108: This appears to be another instance where no translatable text is provided inside `i18n.tr()`.

src/app/DownloadsDialog.qml
```diff
@@ -0,0 +1,190 @@
...
+import Ubuntu.Components 1.3
+import Ubuntu.Components.Popups 1.3
+import QtQuick.Layouts 1.3
+import Ubuntu.Content 1.3
+import webbrowsercommon.private 0.1
+
+import "MimeTypeMapper.js" as MimeTypeMapper
+import "FileUtils.js" as FileUtils
+
+Popover {
```
UniversalSuperBox, line 30: It is very difficult to see where the "Recent downloads" popup begins and the page ends when both have a light background. Is this expected?

image

src/app/ContentExportDialog.qml
```diff
@@ -0,0 +1,120 @@
+/*
```
UniversalSuperBox, line 1: I'm a bit concerned with the size of this PR, both in commits and lines. Is it possible to resubmit it as multiple PRs, each one making a change which can stand without the others?

As it is, I would have difficulty merging this PR without making the history confusing to read for anyone trying to trace issues in the future.

UniversalSuperBox (changes requested):
My primary concerns:

* This PR is really big! Can we split it up?
* There are changes to translatable strings. The translation file templates must be rebuilt and included when this happens.

src/app/ContentExportDialog.qml
kugiigi: I think this is @balcy 's change that was merged into this PR which I can't remember anymore why...I think for rebasing to current xenial 😅

src/app/DownloadsDialog.qml
kugiigi: I can't remember completely anymore why I did this but I think this is because different languages writes percentage differently like for example Arabic it's reversed [https://en.wikipedia.org/wiki/Percent_sign]

src/app/DownloadDelegate.qml
kugiigi: Hmm, the comment is a copy/paste mistake 😅
Can't remember what my reason behind or if this wasn't intended but I think it's because other languages may also display `/s` differently?

src/app/DownloadsDialog.qml
kugiigi: I used the default theme/color for Popover so I guess this is just the result of that. Maybe we can improve this by the changing the background a bit. Not sure yet if we can adjust the shadows of the popover.

src/app/ContentExportDialog.qml
kugiigi: I'll check if I can separate them into smaller PRs but I may only be able to do so after two weeks 😅 But I'll try to do it soon as I can.

src/app/DownloadsDialog.qml
cibersheep: Popover relays on the dark layer underneath :)

UniversalSuperBox:
Closing; This PR has conflicts with the base image and must be resubmitted as more, smaller changes.

mateosalta:
> I was supposed to attach some but apparently I completely forgot 😅
> Here's a couple.
> screenshot20200721_025816857
> screenshot20200721_025825117

I really like the pause button, it fits nice as we move more away from text big buttons in many situations :)

PR #337, opened 2020-07-13
hide chrome while recent view (tab list) is visible

### LINKED ISSUE #325: It's too hard to close the current tab
Reproduction:
* open five tabs and surf somewhere
* try to close the current tab

Observed:
It is quite cumbersome. I have to
* swipe up to the get the tab list 
* select another tab 
* swipe up again
* find the original tab in the list
* hit the x 

Expected:
I would expect that with two gestures and without any searching on the screen I should be able to close the current tab

balcy:
fixes https://github.com/ubports/morph-browser/issues/325

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -551,9 +551,12 @@ Common.BrowserView {
         anchors.fill: parent
         visible: bottomEdgeHandle.dragging || tabslist.animating || (state == "shown")
         onVisibleChanged: {
-            if (visible)
-            {
-                currentWebview.hideContextMenu()
+            if (visible) {
+                currentWebview.hideContextMenu();
+                chrome.state = "hidden";
+            }
+            else {
+                chrome.state = "shown";
             }
         }
 


kugiigi:
Oh it's this simple. So it was a design change and not a workaround to a technical issue?

balcy:
I remember that in the first morph versions in order to close the current tab, you had to switch to another tab, and then close the one you have been on. That's why I added the X left to the main menu opener.
But you are right that it is more consistent if the current tab is part of the tab list.
Hard to say why / when the hiding of the chrome was removed. Maybe a binding got removed accidently ? Did it work in webbrowser-app?

dano6:
I have stolen commit from here in https://github.com/ubports/morph-browser/pull/382

mateosalta:
yep chrome didn't show in tab list in webbrowser-app, I think it got there by accident when transferring to qtwebengine

UniversalSuperBox (changes requested):
Can this be handled with a binding instead of an imperative state machine? (`state: recentView.visible ? "hidden" : "shown"`)

UniversalSuperBox:
This was included in https://github.com/ubports/morph-browser/pull/382

PR #348, opened 2020-07-24
Xenial   nohighdpiautoscale

balcy:


mateosalta:
I think there is a strange thing happening with zoom factor with fullscreen video, i was trying it instead of autoscale because of dropdown bugs.

balcy:
ok this effect is always there if a zoomFactor != 1.0 is used.
In the "no auto scale" branch 100% zoom (set in zoom menu) is in fact 150% (or more depending on the phone).
If you change the zoom factor in the xenial branch, the video is too big in fullscreen mode, too.
So we could always set 1.0 in fullscreen mode.
The problem is, that without (auto) scale, the video looks not good, e.g. like a lower res video

balcy:
> I think there is a strange thing happening with zoom factor with fullscreen video, i was trying it instead of autoscale because of dropdown bugs.

I've opened https://github.com/ubports/morph-browser/issues/360 for it

balcy:
@mateosalta @UniversalSuperBox the latest checkin should correct the video size, but without high DPI scaling the screen has the lower pixel density, so for me the video looks not as nice as with scaling enabled.
(so we have less dpi in this case)

mateosalta:
hm, yeah. not ideal. wonder if there is a way to set the fullscreen width? maybe it thinks the size is bigger because of zoom?

PR #351, opened 2020-07-25
Update WebApp.qml

### LINKED ISSUE #338: QML layouting issue in webapp-container ?
Some webapps created with the webber app may not display the options strip of the website at the bottom of the screen. This is very annoying for the use of these webapps.
This has been observed for a webapp created from https://m.lequipe.fr/Tennis/ when   **--enable-addressbar flag is set.**
But for my bank site, which uses an option strip, even when --enable-addressbar flag is NOT set, this option strip is not visible: impossible to use the bank webapp!
see full discussion here https://github.com/timsueberkrueb/webber/issues/27

balcy:
subtract height of the chrome from the height of the WebApp. (otherwise the bottom is cut off)
fixes https://github.com/ubports/morph-browser/issues/338

### DIFF
--- a/src/app/webcontainer/WebApp.qml
+++ b/src/app/webcontainer/WebApp.qml
@@ -214,7 +214,7 @@ Common.BrowserView {
                 top: chromeLoader.bottom
             }
             // https://github.com/ubports/qtwebengine-opensource-src-packaging/issues/38
-            height: Math.round(parent.height - osk.height)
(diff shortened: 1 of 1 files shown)


UniversalSuperBox (approved):
Tested working on OnePlus One. It's now possible to scroll all the way to the bottom of the page, and the bottom of the scroll bar is not cut off.

PR #352, opened 2020-07-26
Update ZoomControls.qml

### LINKED ISSUE #350: webview.zoomFactor is not always changed on first time
There is QT Bug: https://bugreports.qt.io/browse/QTBUG-84313
During my tests on the "xenial_-_nohighdpiautoscale" branch, I've seen that zoom in / zoom out actions (Zoom menu) are sometimes just ignored, the webview.zoomFactor value was unchanged.

The workaround is to set the zoom factor twice.

### LINKED ISSUE #360: Video too big / too small in fullscreen mode, when zoomFactor != 1.0
- go to a page that has a video in landscape format that can be put to full screen
- device is in landscape orientation

1)
- set the zoom to 50% 
- set video to full screen
expected result: the video covers the whole screen
actual result: video is smaller (covers only part of the screen)

2)
- set the zoom to 150% 
- set video to full screen
expected result: the video covers the whole screen
actual result: video is bigger (only part of the video is displayed on the screen)

balcy:
1)
workaround for zoom changes (https://bugreports.qt.io/browse/QTBUG-84313)
fixes https://github.com/ubports/morph-browser/issues/350

2)
fixes https://github.com/ubports/morph-browser/issues/360

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -702,6 +702,7 @@ WebView {
         }
 
     onFullScreenRequested: function(request) {
+        webview.zoomFactor = 1.0;
         request.accept();
     }
 
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -702,9 +702,14 @@ WebView {
         }
 
     onFullScreenRequested: function(request) {
-        // twice because of QTBUG-84313
-        webview.zoomFactor = 1.0;
-        webview.zoomFactor = 1.0;
+        if (webview.isFullSceeen) {
+          webview.zoomController.refresh();
+        }
+        else {
+          // twice because of QTBUG-84313
+          webview.zoomFactor = 1.0;
+          webview.zoomFactor = 1.0;
+        }
         request.accept();
     }
 
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -702,13 +702,12 @@ WebView {
         }
 
     onFullScreenRequested: function(request) {
-        if (webview.isFullSceeen) {
-          webview.zoomController.refresh();
-        }
-        else {
+        if (request.toggleOn) {
           // twice because of QTBUG-84313
           webview.zoomFactor = 1.0;
           webview.zoomFactor = 1.0;
+        } else {
+          webview.zoomController.refresh();
         }
         request.accept();
     }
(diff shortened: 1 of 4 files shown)


src/app/WebViewImpl.qml
```diff
@@ -699,6 +699,14 @@ WebView {
         }
 
     onFullScreenRequested: function(request) {
+        if (webview.isFullSceeen) {
+          webview.zoomController.refresh();
+        }
+        else {
+          // twice because of QTBUG-84313
+          webview.zoomFactor = 1.0;
+          webview.zoomFactor = 1.0;
```
UniversalSuperBox, line 708: Why does zoom need to be reset to 1 when leaving fullscreen? Is this bypassing zoomController?

src/app/webbrowser/Browser.qml
```diff
@@ -844,16 +844,16 @@ Common.BrowserView {
             target: browser.currentWebview
             onLoadingChanged: {
                 if (browser.currentWebview.loading) {
-                    chrome.state = "shown"
+                    chrome.state = "shown";
                 } else if (browser.currentWebview.isFullScreen) {
-                    chrome.state = "hidden"
+                    chrome.state = "hidden";
```
UniversalSuperBox: These changes, while useful, are not necessary for this PR. It'd be best if they could be removed from the commit they are added, or split out into their own commit so they're at least atomic.

UniversalSuperBox (changes requested):

src/app/WebViewImpl.qml
balcy: actually it is set to 1 when entering fullscreen. (see https://github.com/ubports/morph-browser/issues/360)
If the current state is fullscreen, it is about to leave fullscreen mode.

Maybe the following would improve the readability ? 
```
onFullScreenRequested: function(request) {
        if (request.toggleOn) {
            // twice because of QTBUG-84313
            webview.zoomFactor = 1.0;
            webview.zoomFactor = 1.0;
        }
        else {
            webview.zoomController.refresh();
        }
        request.accept();
    }
```

balcy: the zoom factor of the zoom controller remains unchanged, only in fullscreen mode the browser zoom goes to 1.0
So it is bypassed while in fullscreen mode

src/app/webbrowser/Browser.qml
balcy: luckily it was like the example in https://stackoverflow.com/a/9103397 : )
So I've removed the changes from the loadingChanged function.

UniversalSuperBox (approved):
Looks good and resolves the listed issues.

PR #353, opened 2020-07-26
remove the setting of QTWEBENGINE_CHROMIUM_FLAGS, because they are al…

balcy:
…ready set outside of morph-browser / webapp-container

setting them here can lead to situations where morph-browser / webapp-container do work, but other apps not, and any changes have to be made at several locations
see https://github.com/ubports/lxc-android-config/blob/xenial/etc/profile.d/disable-qtwebengine-gpu.sh

- undo the roundings

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -361,8 +361,7 @@ Common.BrowserView {
                 top: parent.top
             }
 
-            // https://github.com/ubports/qtwebengine-opensource-src-packaging/issues/38
-            height: Math.round(parent.height - osk.height - bottomEdgeBar.height)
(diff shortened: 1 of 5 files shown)


mateosalta:
I have been experimenting with several chromimum flags with morph. will this remove the ability to use them with morph? Things like security settings, and scroll settings, that a user might want to set at a browser level. also have been testing diffrent gpu flags on the pinephone.

balcy:
no, they have only been set by morph, if the QTWEBENGINE_CHROMIUM_FLAGS was empty. Additionally they are set in https://github.com/ubports/lxc-android-config/blob/xenial/etc/profile.d/disable-qtwebengine-gpu.sh
So your own flags would override them anyway.

UniversalSuperBox (approved):

PR #365, opened 2020-08-09
update addressbar.text via timer

### LINKED ISSUE #299: URL input fields becomes unresponsive while searching hints
Whenever I enter an URL in the top URL inout field I kinda feel blocked by the browser since it starts a search immediately and becomes unresponsive until the result with the hints is back. To solve this either:
- delay the start of the search until the user stopped typing a bit 
AND/OR
- search in a separate thread, that can be stopped/interrupted/restarted when new characters are entered by the user (again with some delay)
The whole feature becomes pointless if I always have to wait until my 3 or 4 characters lead to a result I cannot use ;)

balcy:
fixes https://github.com/ubports/morph-browser/issues/299

### DIFF
--- a/src/app/webbrowser/AddressBar.qml
+++ b/src/app/webbrowser/AddressBar.qml
@@ -27,7 +27,7 @@ FocusScope {
 
     property alias icon: favicon.source
     property bool incognito: false
-    property alias text: textField.text
+    property string text
     property bool bookmarked: false
     signal toggleBookmark()
     property url requestedUrl
@@ -73,11 +73,19 @@ FocusScope {
         value: findInPageMode ? textField.text : ""
     }
 
+    Timer {
+        id: searchTextTimer
+        onTriggered: (addressbar.text = textField.text)
+        interval: 900
+    }
+
     TextField {
         id: textField
         objectName: "addressBarTextField"
 
         anchors.fill: parent
+
+        onTextChanged: searchTextTimer.restart()
 
         primaryItem: Item {
             id: icons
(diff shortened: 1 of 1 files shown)


src/app/webbrowser/AddressBar.qml
```diff
@@ -467,68 +476,79 @@ FocusScope {
...
     onFindInPageModeChanged: {
-        if (findInPageMode) return
-        if (canSimplifyText) {
-            text = internal.simplifyUrl(actualUrl)
-            internal.simplified = true
+        if (findInPageMode) {
+            textField.text = "";
+        } else if (canSimplifyText) {
+            textField.text = internal.simplifyUrl(actualUrl);
+            addressBar.text = textField.text;
```
UniversalSuperBox: This line causes the following warning in the log:

```
file:///usr/share/morph-browser/webbrowser/AddressBar.qml:546: ReferenceError: addressBar is not defined
```

UniversalSuperBox (changes requested):

UniversalSuperBox (commented):
Gave this another try today -- right now, the search results from the last text entry pause stay up until the next pause. Might it be better for the results to go away immediately when the user starts typing with the delay to perform the search kept?

lduboeuf:
hi, any news ?

UniversalSuperBox:
https://github.com/ubports/morph-browser/issues/299#issuecomment-840158054

It's unclear who experiences this issue, and those who do have not verified that this fixes it.

lduboeuf:
will give  a try, i thought it was solving the issue at Morph or webapp startup where the address field focus is stealed by the webpage

balcy:
at that time it was possible for me to reproduce the issue:
- the phone should have a very long browser history (e.g. not cleared for a long time)
- for each keypress in the address bar there is a small delay, because the history is searched for suggestions (previously visited pages)
- with the fix you can type in the address bar without the delay

in my opinion it cannot have any effect no webapps, because they use another address bar (where you cannot type), and is not related to keyboar focus / startup problems, but only the typing in the address bar text field.  (see linked issue 299)

Issue #347, opened 2020-07-24
Devel: Can't stay authenticated 


lduboeuf:
in apps that use the Morph.Web component, we can't stay connected.
Tested on Gmail, Youtube, Twitter webapps
Is there anything to be done ?

balcy:
I think the profile of the new QtWebEngine (https://doc.qt.io/qt-5/qml-qtwebengine-webengineprofile.html) has now `offTheRecord:true` as default. 
https://bugreports.qt.io/browse/QTBUG-79685
So setting `context.offTheRecord:true` in the MorphWebView should solve it. Not sure if we can / should sth. globally, because incognito by default seems better privacy wise.

mateosalta:
nothing is staying signed in in the browser now...

balcy:
gitlab / github stays logged in for me in morph, and github in the webapp, too.

mretallack:
Hi,
Is there a run-time work around for setting offTheRecord to false?

mretallack:
I have found that setting offTheRecord to False in MorphWebContext allows the cookies to persist. I have no idea if this is the correct solution.

 I have attached a Patch


[fix-offTheRecord.patch.txt](https://github.com/ubports/morph-browser/files/4975363/fix-offTheRecord.patch.txt)

Update: This may only effect PinePhone, https://github.com/ubports/morph-browser/blob/bb9f86efd727f100e1cc60390db0955521c02240/src/Morph/Web/MorphSharedWebContext.qml#L27

balcy:
yes, but the current xenial (DEV branch) version does include this change already.

https://github.com/ubports/morph-browser/blob/xenial/src/Morph/Web/MorphSharedWebContext.qml

Is it possible morph-browser does not use the xenial branch on some phones / channels (e.g. pinephone) ?

balcy:
oh there are https://github.com/ubports/morph-browser/tree/xenial_-_edge and https://github.com/ubports/morph-browser/tree/xenial_-_edge_-_pine branches...

balcy:
in the QA group I've seen Martin had the following in the log: `qml: the temporary default profile does no longer exit`
Therefore I've changed the way the temporary profile (to get the chromium version) is created.
It would not explain that no sessions at all are saved though.
https://github.com/ubports/morph-browser/pull/359

balcy:
Is it still the case that the one can't stay authenticated with the current dev version ?

lduboeuf:
[...]

PR #376, opened 2020-09-01
Set a default for offTheRecord in MorphWebContext

UniversalSuperBox:
QtWebEngine 5.13 changed the default value of offTheRecord from false to true. This means that apps will dump all cookies and site data on exit by default.

If someone is using the MorphWebContext as part of their app, they expect the old behavior with offTheRecord = false. So we'll just set that.

https://github.com/ubports/morph-browser/issues/347

This does not help apps which are using WebEngineProfile directly without setting offTheRecord. Those apps will need to set "offTheRecord: false" on their Profile by themselves.

### DIFF
--- a/src/Morph/Web/MorphWebContext.qml
+++ b/src/Morph/Web/MorphWebContext.qml
@@ -28,6 +28,8 @@ WebEngineProfile {
     property alias incognito: oxideContext.offTheRecord
     readonly property string defaultUserAgent: __ua.defaultUA
 
+    offTheRecord: false
+
     dataPath: dataLocation
 
     cachePath: cacheLocation


UniversalSuperBox:
https://github.com/qt/qtwebengine/commit/6e28e764d7c425f942820f0881231eb83e553e6e shows exactly when this change occurred. This may not be a fix.

UniversalSuperBox:
This is, in fact, a fix. offTheRecord should be set automatically if a dataPath (persistentStoragePath) is provided, but if it is not provided *and* offTheRecord is set manually to false the StandardPath is used instead: 

https://github.com/qt/qtwebengine/blob/08375ee796a545e3916feb24e0b35d561e6f0049/src/core/profile_adapter.cpp#L246-L259

mariogrip (approved):
LGTM :+1:

UniversalSuperBox:
While https://github.com/ubports/qtwebengine-opensource-src-packaging/pull/41 was merged and fixed the issue instead, I believe it's still worth it to merge this to make API breaks inside Morph.Web explicit.

PR #378, opened 2020-09-08
pinephone fullscreen fix, wayland shell integration

mateosalta:
wayland shell integration for fullscreen that hides indicators


dobey (approved):
Looks OK. Thanks!

PR #382, opened 2020-09-17
Implement swipe to close tab and fix tab previews

### LINKED ISSUE #119: Morph Browser tab website preview only on selected tab
- Device: Nexus 5
- Channel: 16.04 RC
- Build: 2018-W48

### Steps to reproduce
Open at least two sites in tabs in Morph Browser, open tab selection menu.

### Expected behavior
Website preview is expected to be displayed in all tabs.

### Actual behavior
Only previously selected tab shows site preview, the rest are empty. The other tabs does not seem to be unloaded because when selected the website is shown instantly.

screenshot20181130_023701644

### LINKED ISSUE #325: It's too hard to close the current tab

dano6:
This should fix https://github.com/ubports/morph-browser/issues/119 and fix https://github.com/ubports/morph-browser/issues/325

### DIFF
--- a/src/app/webbrowser/BrowserTab.qml
+++ b/src/app/webbrowser/BrowserTab.qml
@@ -45,10 +45,10 @@ FocusScope {
     property bool incognito
     readonly property bool empty: !url.toString() && !initialUrl.toString() && !restoreState && !request
     property bool loadingPreview: false
-    readonly property size previewSize: Qt.size(webview.width*Screen.devicePixelRatio,
-                                                webview.height*Screen.devicePixelRatio)
-    readonly property size previewThumbnailSize: Qt.size(webview.width/1.5,
-                                                         webview.height/1.5)
+    readonly property size previewSize: webview ? Qt.size(webview.width*Screen.devicePixelRatio,
+                                                webview.height*Screen.devicePixelRatio) : Qt.size(0,0)
+    readonly property size previewThumbnailSize: webview ? Qt.size(webview.width/1.5,
+                                                         webview.height/1.5) : Qt.size(0,0)
     //store reference to preview to avoid clearing by garbage collector
     property var previewCache
     visible: false
--- a/src/app/webbrowser/BrowserTab.qml
+++ b/src/app/webbrowser/BrowserTab.qml
@@ -201,33 +201,15 @@ FocusScope {
         }
     }
 
-    // Take a capture of the current page shortly after it has finished
-    // loading to give rendering an opportunity to complete. There is
-    // unfortunately no signal to notify us when rendering has completed.
-    Timer {
-        id: delayedCapture
-        interval: 500
-        onTriggered: {
-            if (recentView.visible)
-                return
-            if (webview && current && visible && !internal.hiding) {
+    Connections {
+        target: recentView
+        onVisibleChanged: {
+            if(visible && current && !empty && !webview.incognito) {
+                preview = ""
                 webview.grabToImage(function(result) {
                     PreviewManager.saveToDisk(result, url)
-                },Qt.size(webview.width*Screen.devicePixelRatio,webview.height*Screen.devicePixelRatio))
+                },Qt.size(webview.width*Screen.devicePixelRatio,webview.height*Screen.devicePixelRatio));
             }
-        }
-    }
-    Connections {
-        target: incognito ? null : webview
-        onLoadingChanged: {
-            if (!webview.loading)
-                delayedCapture.restart()
-        }
-        onScrollPositionChanged: {
-            delayedCapture.restart()
-        }
-        onWidthChanged: {
-            delayedCapture.restart()
         }
     }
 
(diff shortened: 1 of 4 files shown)


dano6:
For now we need save preview image to disk, but QT 5,13 brings keeping it just in memory https://doc.qt.io/qt-5.14/qml-qtquick-itemgrabresult.html. Snapshoting of current page after every scroll currently makes scrolling a bit junky, so we snapshoting it just after loading of page ended.

fredldotme:
I've tested this now. Good for a start but there are two minor issues.

- When opening the tab overview it sometimes shows the capture of an old scroll state on the topmost tab.
- Also, I'd like to see some kind of transition animation between normal browsing and tab overview. This would improve the overall UX and hide from the fact that the chrome now disappears when changing views.

dano6:
@fredldotme  Please try now, I improved snapshooting of first tab and added animation for chrome showing and hiding

fredldotme:
There's still some oddity when opening the tab spread whenever the active tab is loading.

dano6:
@fredldotme please try now, I tried to do it differently.  There is possibility that the first tab would be webview, but swiping of that would be almost impossible.

fredldotme:
This is getting good! The delays for taking the snapshots are mostly gone.

I noticed though that the opacity animation when bringing up the TabsList is making it look as if the currently active tab is blinking. Do you think you could improve that a little?

dano6:
I changed animation and downscaled a bit previews. Now it seems to be smoother and more responsive

fredldotme:
Yeah I noticed, good job! I wouldn't have noticed the downscaling if you hadn't mentioned it.

Please take a look at the inline comments I made, they would make the PR better and more "approvable".

dano6:
@fredldotme 
> Please take a look at the inline comments I made, they would make the PR better and more "approvable".

Sorry, I don't see any comments, can you please check it?

fredldotme:
In src/app/webbrowser/BrowserTab.qml, lines 201 & 213, please make sure to separate these values into meaningfully named variables/constants, so that it's readable again.

fredldotme:
This latest change seems to have broken the previews after a morph-browser restart. Though not functional before the PR, it would be nice to keep having it's functionality to make it in time for the next OTA.

dano6:
I fixed restoring previews from storage after browser restart. We are doing now two previews: Hi-res - stored only in memory, and Low-res - stored on disk. Storing High resolution previews in storage make UI bit choppy and it is not healthy for the hardware.  https://github.com/ubports/morph-browser/pull/382#issuecomment-697166426 is not true, that functionality was already there, but not in documentation. I think this MR is now ready for review, I do not see more space for improvement.

src/app/webbrowser/BrowserTab.qml
```diff
@@ -44,25 +44,19 @@ FocusScope {
     readonly property real lastCurrent: internal.lastCurrent
     property bool incognito
     readonly property bool empty: !url.toString() && !initialUrl.toString() && !restoreState && !request
+    property bool loadingPreview: false
+    readonly property size previewSize: Qt.size(webview.width*Screen.devicePixelRatio,
+                                                webview.height*Screen.devicePixelRatio)
+    readonly property size previewThumbnailSize: Qt.size(webview.width/1.5,
+                                                         webview.height/1.5)
```
UniversalSuperBox: The following warnings occur upon closing a tab:

```
file:///usr/share/morph-browser/webbrowser/BrowserTab.qml:48: TypeError: Cannot read property 'width' of null
file:///usr/share/morph-browser/webbrowser/BrowserTab.qml:50: TypeError: Cannot read property 'width' of null
file:///usr/share/morph-browser/webbrowser/BrowserTab.qml:48: TypeError: Cannot read property 'width' of null
file:///usr/share/morph-browser/webbrowser/BrowserTab.qml:50: TypeError: Cannot read property 'width' of null
```

I know the tab is closing anyway and these warnings probably don't matter, but I prefer not adding new ones.

src/app/webbrowser/BrowserTab.qml
```diff
@@ -190,40 +184,36 @@ FocusScope {
...
+                preview = result.url
+                previewCache = result
+            },previewSize);
 
+            //save previews to disk for newtabpage and tab during grabbing
+            webview.grabToImage(function(result) {
+                internal.hiding = false
                 PreviewManager.saveToDisk(result, url)
-            })
+            },previewThumbnailSize);
```
UniversalSuperBox, line 196: I'm unsure where it comes from, but the following warning is thrown when tabs become visible:

```
Ignoring sourceSize request for image url that came from grabToImage. Use the targetSize parameter of the grabToImage() function instead.
```

UniversalSuperBox (changes requested):

src/app/webbrowser/BrowserTab.qml
dano6: This is coming from depths of QT when HIDPI is used. Probably QT bug, there is chance that it will fix the new QT. I am using targetSize parameter, not sure why this warning.

src/app/webbrowser/BrowserTab.qml
dano6: Fixed.

UniversalSuperBox (approved):
Alright, all looks good then. Let's do it. :sunglasses:

PR #390, opened 2020-09-26
allow pasting to clipboard from javascript

### LINKED ISSUE #389: Enable copying content to the Clipboard via JavaScript
It is not always possible to copy text to the clipboard via Javascript.

- open https://github.com/ubports/morph-browser/blob/xenial/morph-browser.png in morph
- click the dots right (not the ones in the header bar, but the one for the file)
- click copy path
- paste the path e.g. in the terminal app, or the address bar
image 

- expected result: the string `morph-browser.png` is pasted
- actual result: the previous content of the clipboard (or nothing) is pasted 

setting https://doc.qt.io/qt-5/qml-qtwebengine-webenginesettings.html#javascriptCanAccessClipboard-prop to true solves the issue

balcy:
fixes https://github.com/ubports/morph-browser/issues/389

### DIFF
--- a/src/app/WebViewImpl.qml
+++ b/src/app/WebViewImpl.qml
@@ -61,6 +61,9 @@ WebView {
     // furthermore pages saved as PDF would not be downloaded but only displayed as PDF
     settings.pdfViewerEnabled: false
 
+    // allow pasting to clipboard from javascript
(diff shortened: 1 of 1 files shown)


mariogrip (approved):
LGTM :+1:

PR #394, opened 2020-10-05
tab look changes

mateosalta:
extend tab downward for full connected tab look (removes corner artifacts), remove tab outline for a more flat look and rely on tab shadow.


mateosalta:
screenshot20201005_143809453
screenshot20201005_143549393

UniversalSuperBox:
Hi @mateosalta, it appears that this change conflicts with some changes that have been made in the xenial branch. Could you update your branch?

UniversalSuperBox (changes requested):
I don't think the divider should be removed. If two pages have similar colors, there is no separation between them.

Tabs without divider

Whereas with the divider, the pages look separate from each other:

Tabs with divider

mateosalta:
Oh, yes i can add that back in. I was trying to fix the slight drop shadow instead but was not having luck :) will focus on the tab part for this one

mateosalta:
Awesome @UniversalSuperBox I have added the divider back, now this pr only changes the tab look, specifically: outline of tab, shape of the bottom of the tab extends downward so that it is more like a tab similar to the tabs in morph's wider view.
photo_2021-05-18_18-08-10
photo_2021-05-18_18-08-01

UniversalSuperBox (approved):

PR #395, opened 2020-10-06
embeded switches to enable better scrollbars

mateosalta:
enable the arguments list, and use --enable-embeded-switches, this has the primary effect of showing very nice overlay scrollbars


UniversalSuperBox:
@dobey, in cases like this, does it make more sense to put this argument in the desktop file or somewhere else, rather than inserting it in code?

mateosalta:
Ah, I didn't think of that, since it isn't a chromimum flag I think that would work, will test and see.

UniversalSuperBox:
Superseded by https://github.com/ubports/morph-browser/pull/407

PR #397, opened 2020-10-11
Set Desktop Mode from Main Menu

### LINKED ISSUE #324: Feature: "request desktop version"
Firefox on Android has a very helpful menu: "Request desktop site"

You can tick it in a tab and then the website switches from mobile view to desktop view. For example on github. I find it to be a very helpful for misbehaving websites. E.g., things that are not (easily) possible on a mobile site. 

NB: Yes, some websites have their own toggles to switch to desktop mode, but then I sometimes had problems finding a way to switch back.

niutech:
This small addition shows the _Set Desktop Mode_ checkable action in the Main / Hamburger Menu, so that users can easily switch between the desktop and mobile versions of the web page. Fixes #324.

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -818,6 +818,14 @@ Common.BrowserView {
                 iconName: "zoom-in"
                 enabled: currentWebview && (currentWebview.url.toString() !== "")
                 onTriggered: currentWebview.showZoomMenu()
+            },
+            Action {
+                objectName: "desktopmode"
+                text: i18n.tr("Set Desktop mode")
+                iconName: "computer"
+                checkable: true
+                checked: settings.setDesktopMode
+                onToggled: settings.setDesktopMode = checked
             }
 
         ]


balcy:
If it's at the bottom of the menu I think it's fine to add it in my opinion. (It can be ignored by users who not need it)

kugiigi:
In my opinion, this actually deserves to be in the menu more than some of the current ones like "Save / Print to PDF".
If the menu is versatile, it would be ideal to rearrange the menu into maybe sub-menus and actual controls like for zoom 😄

balcy (approved):

UniversalSuperBox (changes requested):
Hey, that's really neat! Thank you! This is a little strange to use, though...

This menu item does not change to reflect whether desktop mode is on or off. I think the icon and/or text should change to reflect what you're about to set (assuming you can't put an actual checkbox there). Check out our list of icons to see some possible candidates:

https://docs.ubports.com/projects/icons/en/latest/

I might suggest the text being "Set desktop mode" and "Set mobile mode" or something more clear, and the icons being `phone-smartphone-symbolic` and `computer-symbolic` for which you would set if you tapped the icon.

Might also be a good idea to only show the action if the device is small enough to actually switch between the two modes!

niutech:
@UniversalSuperBox I can make these changes, but why can't we use a generic [checkable action](https://doc.qt.io/qt-5/qml-qtquick-controls2-action.html#checkable-prop) with a checkbox like this:
checkable action

UniversalSuperBox:
If that works, yes, just do that :)

PR #407, opened 2020-11-24
enable embeded switches, in desktop file

mateosalta:
simpler to add the switch in the exec line of the desktop, then adding complicated code :)

### DIFF
--- a/src/app/webbrowser/morph-browser.desktop.in.in
+++ b/src/app/webbrowser/morph-browser.desktop.in.in
@@ -6,7 +6,7 @@ _Comment=Browse the World Wide Web using QtWebEngine
 _Keywords=Internet;WWW;Browser;Web;Explorer
 Type=Application
 Icon=@CMAKE_INSTALL_FULL_DATADIR@/morph-browser/morph-browser.png
-Exec=morph-browser %u
+Exec=morph-browser %u --enable-embeded-switches
 Terminal=false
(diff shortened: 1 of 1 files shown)


UniversalSuperBox:
Seems like these scrollbars can't be touched and dragged for fast scrolling, is that correct?

mateosalta:
I think so, points on the old scroll bars:
- old ones were too small to detect touch anyway
- they caused the page to move over and didn't overlay without zooming
- they didn't disappear once scrolling stopped
- relied on user scrip injection, not ideal

If the old ones did respond to mouse interaction, perhaps we could move back to the in code method and come up with a "if desktop, windowed mode" type of switch?

right now I see more benefits for these new ones on mobile.

kugiigi:
I think the scrollbar needs to be interactive all the time. It'll be hard to detect mobile or desktop since you can still have a mouse while in narrow layout. Also, the scrollbar is too thin even for mouse use at the moment.
What I want to try as a solution is to add hover and scrolling style changes to the scroll bar via javascript.

mateosalta:
they need a lot of work :) I need to test but I have a suspicion that maybe our js injection might be causing the scroll lock up, I think that bug started when the injection point of the scrip was changed.

mateosalta:
I found some better flags! so i should submit those to the android config file?

PR #412, opened 2020-12-02
Delete action actually deletes downloaded file

### LINKED ISSUE #330: Deleting a download does not delete the file

kugiigi:
Fixes #330 
Originally from #336

### DIFF
--- a/src/app/DownloadDelegate.qml
+++ b/src/app/DownloadDelegate.qml
@@ -224,7 +224,7 @@ ListItem {
                 enabled: error.visible || !incomplete
                 onTriggered: {
 
-                   cancelled()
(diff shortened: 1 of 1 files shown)


dobey (approved):
Looks good, thanks!

Issue #311, opened 2020-04-12
Crash when adding favorite


Keneda-ut:
Meizu MX4 OTA-11 stable

If i tap on the star in url field to add the website to favorites, morph shuts down.
Next morph start, the website is in favorites, and if i tap star again, then website is took off favorite without crashing morph.
But next time i try to favorite it crash again, etc...

Way to reproduce : 

- Install/launch UT Tweak Tool, go scaling option, make grid size 20

- Launch Morph and Try to add favorite

Tried from 18 to 24 (default is 23), only occurs with grid size at 20.

Keneda-ut:
OK this bug is related with UT Tweak Tool app :
https://forums.ubports.com/topic/4072/morph-browser-crash-need-help-feedback/5

But i don't know if it's morph or UT Tweak Tool to be "patched".

Keneda-ut:
Opened issue on UTTT gitlab : https://gitlab.com/myii/ut-tweak-tool/-/issues/42
Bug confirmed by other MX4 user.

cibersheep:
Confirmed on Arale, Dev Channel.

Setting scalation to 20, makes Morph crash when adding a website to favorites

myii:
Collected the error logs for #311 and #312 at https://gitlab.com/myii/ut-tweak-tool/-/issues/42#note_322398115.  Both are almost identical.  Both actions involve pop-ups.  This is the log for favourites:

```
file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Components/1.3/OptionSelector.qml:330:13: QML ListView: Binding loop detected for property "itemHeight"
propsReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.300\" (uid=32011 pid=557 comm=\"morph-browser \") interface=\"org.freedesktop.DB
us.Properties\" member=\"GetAll\" error name=\"(unset)\" requested_reply=\"0\" destination=\"org.freedesktop.NetworkManager\" (uid=0 pid=2163 comm=\"NetworkManager \")"
nmReply "An AppArmor policy prevents this sender from sending this message to this recipient; type=\"method_call\", sender=\":1.300\" (uid=32011 pid=557 comm=\"morph-browser \") interface=\"org.freedesktop.Netwo
[...]

PR #414, opened 2020-12-02
Moved content picker into a popover

kugiigi:
Originally from #336 

- Put the content picker when opening downloads into a popup for a more "convergent" design.
- Also moved here the open in browser action buttons

### DIFF
--- a/src/app/DownloadsDialog.qml
+++ b/src/app/DownloadsDialog.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Canonical Ltd.
+ * Copyright 2021 UBports Foundation
  *
  * This file is part of morph-browser.
  *
@@ -117,20 +117,19 @@ Popover {
                 icon: MimeDatabase.iconForMimetype(modelData.mimeType)
                 
                 onClicked: {
-                    /* TODO: Enable once content picker in a popover is merged  */
-                    /*if (!incomplete && !error) {
+                    if (!incomplete && !error) {
                         var properties = {"path": download.path, "contentType": MimeTypeMapper.mimeTypeToContentType(download.mimeType), "mimeType": download.mimeType, "downloadUrl": download.url}
                         var exportDialog = PopupUtils.open(Qt.resolvedUrl("ContentExportDialog.qml"), downloadsDialog.parent, properties)
                         exportDialog.preview.connect(downloadsDialog.preview)
-                    } else {*/
-                    if (download) {
-                        if (paused) {
-                            download.resume()
-                        } else {
-                            download.pause()
+                    } else {
+                        if (download) {
+                            if (paused) {
+                                download.resume()
+                            } else {
+                                download.pause()
+                            }
                         }
                     }
-                    //}
                 }
                 
                 onRemove: downloadsListView.removeItem(index)
--- a/src/app/DownloadsDialog.qml
+++ b/src/app/DownloadsDialog.qml
@@ -41,7 +41,6 @@ Popover {
     property real preferredHeight:  downloadsDialogColumn.height + units.gu(2)
     
     signal showDownloadsPage()
-    signal preview(string url)
 
     contentHeight: preferredHeight > maximumHeight ? maximumHeight : preferredHeight
     contentWidth: preferredWidth > maximumWidth ? maximumWidth : preferredWidth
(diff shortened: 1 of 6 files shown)


UniversalSuperBox (changes requested):
This appears to introduce a warning when the content picker is opened by tapping on a download:

```file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Content/ContentPeerPicker13.qml:196: TypeError: Type error```

We should investigate why this happens, and if possible fix it, before merging.

**Comment by kugiigi** [comment]
> This appears to introduce a warning when the content picker is opened by tapping on a download:
> 
> `file:///usr/lib/arm-linux-gnueabihf/qt5/qml/Ubuntu/Content/ContentPeerPicker13.qml:196: TypeError: Type error`
> 
> We should investigate why this happens, and if possible fix it, before merging.

I can't find exactly what's causing the error so I decided to just fix those annoying `TypeError` errors in content hub :grinning:  
ubports/content-hub#17

I've also rebased it and fixed the conflicts. I also enabled the commented out codes for opening files from the downloads dialog.
I hope it's fine otherwise, sorry :sweat_smile:

File: src/app/DownloadsDialog.qml [hunk 1]
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015-2016 Canonical Ltd.
+ * Copyright 2021 UBports Foundation
```
UniversalSuperBox, line 2: Whoops... In this case, both copyright statements would be kept. The code was under copyright by Canonical, and now we've added some stake with our changes.

UniversalSuperBox (changes requested):

src/app/DownloadsDialog.qml
kugiigi: No, that file was created in #415 
i actually noticed that there are a few others that we missed in that PR :sweat_smile: 
I'm planning to correct them in a separate PR but I just corrected this one here since I modified it anyway.

But this is definitely a lesson learned for me moving forward - don't just copy paste copyright texts :grinning:

kugiigi:
I changed the dialog from UITK to QQC2 since it's more flexible with placement and sizing.
And it also fixes the issue with UITK popup's visibility and avoid the crash issue on specific scaling (#311)

This also sets the qt version to 5.12 because it was needed to set the dimming background of the popup.

UniversalSuperBox:
Here is a preview of this change at different scale levels

OnePlus One portrait
image

OnePlus One landscape
image

Nexus 7 portrait
image

Nexus 7 landscape
image

Nexus 7 windowed (default scaling)
image

Nexus 7 windowed (small scaling)
image

cibersheep:
Oh. This is a improvement.
Is nice to have the filename in the header and that the header takes less space.

I see the dialog on portrait is anchored to the bottom. They should center vertically as well.

Question (out of curiosity): Why did you choose a Dialog instead of a Page?
Question (might be off topic): What's the difference between the header actions folder and external link? The file is open in the browser whatever I choose.

kugiigi:
> I see the dialog on portrait is anchored to the bottom. They should center vertically as well.

I can't remember exactly the logic when it anchors at the bottom but I think it only happens at certain height and width and not just when in portrait. Otherwise, the dialog is centered (see last picture). I designed it like that so that it's easily reachable. It's a bit similar with android's dialog when opening a file or an intent.

> Question (out of curiosity): Why did you choose a Dialog instead of a Page?

A dialog can be positioned more freely and can be designed with convergence in mind. The content picker doesn't need to be always displayed full especially on large screens.

> Question (might be off topic): What's the difference between the header actions folder and external link? The file is open in the browser whatever I choose

These are existing actions that were not implemented on this PR. In my understanding, one opens the local copy of the file in the browser while the other opens the original dowload link of the file.

cibersheep:
> > I see the dialog on portrait is anchored to the bottom. They should center vertically as well.
> 
> I can't remember exactly the logic when it anchors at the bottom but I think it only happens at certain height and width and not just when in portrait. Otherwise, the dialog is centered (see last picture). I designed it like that so that it's easily reachable. It's a bit similar with android's dialog when opening a file or an intent.

Dialogs in UT are centered horizontally and vertically. Moprh being a core app should follow the same UI design.

Thanks for the other answers :)

kugiigi:
Then maybe we should change that design 😝
I'll change this when I get the time. Unforetunately, I'm not sure when that would be 😅

kugiigi:
I've changed the behavior of the popup. It's now full screen for most phones, not anchored at the bottom.

cibersheep:
There is a dancing sheep somewhere :)

Fuseteam:
what is blocking this PR :3

PR #415, opened 2020-12-03
Downloads UX Changes

### LINKED ISSUE #193: Annoying downloader

kugiigi:
Originally from #336 

- Fixes #193 
- Implemented recent downloads dialog in bot the browser and webapp. This is displayed when a new download started instead of the downloads page.
- A downloads button is added in the navigation bar which wiggles and turns blue when a download completes. In webapps without the title bar, the recent downloads dialog is displayed instead
- Only display the file names in the Downloads page instead of full path.
- Added info such as file size and download speed in the Downloads page
- Added actions such as copy download link and remove from history

### DIFF
--- a/src/app/DownloadsPage.qml
+++ b/src/app/DownloadsPage.qml
@@ -16,14 +16,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import QtQuick 2.4
-import QtWebEngine 1.5
+import QtQuick 2.9
+import QtWebEngine 1.10
 import Ubuntu.Components 1.3
+import Ubuntu.Components.Popups 1.3
 import Ubuntu.Content 1.3
 import webbrowsercommon.private 0.1
 
 import "MimeTypeMapper.js" as MimeTypeMapper
-import "UrlUtils.js" as UrlUtils
+import "FileUtils.js" as FileUtils
 
 BrowserPage {
     id: downloadsItem
@@ -117,7 +118,7 @@ BrowserPage {
         },
         Action {
             iconName: "edit"
-            visible: !selectMode && !pickingMode && !exportPeerPicker.visible
+            visible: !selectMode && !pickingMode
             enabled: downloadsListView.count > 0
             onTriggered: {
                 selectMode = true
@@ -207,9 +208,11 @@ BrowserPage {
         }
 
         delegate: DownloadDelegate {
+            id: downloadDelegate
+                         
             download: ActiveDownloadsSingleton.currentDownloads[model.downloadId]
             downloadId: model.downloadId
-            title: getDisplayPath(model.path)
+            title: FileUtils.getFilename(model.path)
             url: model.url
             image: model.complete && thumbnailLoader.status == Loader.Ready 
                                   && (model.mimetype.indexOf("image") === 0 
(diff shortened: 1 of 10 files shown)


UniversalSuperBox (changes requested):
This PR appears to conflict on DownloadDelegate.qml... Could you rebase it and resolve that conflict? At the moment it isn't testable due to the Oxide removal...

kugiigi:
@balcy Is your commit already what Dalton requested? Do I still need to do something here?

balcy:
yes it was the merge conflict

src/app/DownloadsPage.qml
```diff
@@ -16,14 +16,15 @@
...
-import QtWebEngine 1.5
+import QtQuick 2.9
+import QtWebEngine 1.10
 import Ubuntu.Components 1.3
+import Ubuntu.Components.Popups 1.3
 import Ubuntu.Content 1.3
 import webbrowsercommon.private 0.1
 
 import "MimeTypeMapper.js" as MimeTypeMapper
-import "UrlUtils.js" as UrlUtils
```
UniversalSuperBox, line 26: This PR breaks the ability to open PDF files in the browser directly. The error when tapping the button is:

`file:///usr/share/morph-browser/DownloadsPage.qml:132: ReferenceError: UrlUtils is not defined`

src/app/SimpleDownloadDelegate.qml
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright 2015-2016 Canonical Ltd.
```
UniversalSuperBox: Files that you're creating should at least have a current copyright date, either for yourself or for the UBports Foundation if you wish to assign your copyright to us.

src/app/SimpleDownloadDelegate.qml
```diff
@@ -0,0 +1,164 @@
...
+    
+    divider.visible: false
+
+    MimeData {
+        id: linkMimeData
+        
+        text: download.url
+    }
+    
+    ListItemLayout {
```
UniversalSuperBox, line 50: It doesn't appear that you can tap on downloads in this view to open them. I tried with a PDF file. Is that expected?

UniversalSuperBox (changes requested):
This seems to work well on tablet and mobile usecases, but I've found a few issues.

src/app/DownloadsPage.qml
kugiigi: Oh you are right, I forgot to remove those codes. I guess because I'm suppose to move them to the content picker dialog in #414

src/app/SimpleDownloadDelegate.qml
kugiigi: Apparently, I commented out the codes for this because it needs #414 to be merged first. It uses the same popup :sweat_smile:

src/app/DownloadsPage.qml
kugiigi: I've restored this. I think #414 removes it anyway.

src/app/SimpleDownloadDelegate.qml
kugiigi: I've updated it to UBports and also to 2021

UniversalSuperBox (approved):

PR #416, opened 2020-12-07
Added reopen tab in the toolbar

kugiigi:
This adds a button in the toolbar to reopen the recently closed tab.
This has been available via `Ctrl + Shift + W/T` but it's not accessible in touch devices.


UniversalSuperBox (approved):
Yes indeed, that is a button that reopens a recently closed tab!

PR #417, opened 2020-12-07
Focus the address bar on a new/blank tab even without physical keyboard

kugiigi:
The current implementation is that the address bar is only focused when a physical keyboard is present. With this change, it will always be focused regardless.

I vote for this change because it's a lot more frequent that a user types a new URL/search text when opening a new tab than opening from item in the blank tab (top sites, bookmark, etc.). 

The trade-off is also equal for both cases.
Don't focus
 - Tap on the bookmark/top site
 - Tap on the address bar at the top and start typing

Focus
 - Close the keyboard and tap the bookmark/top site
 - Start typing

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -1299,7 +1299,7 @@ Common.BrowserView {
                 tab.load()
             }
             if (!url.toString()) {
-                maybeFocusAddressBar()
+                focusAddressBar()
             }
         }
 
@@ -1414,7 +1414,7 @@ Common.BrowserView {
                     recentView.focus = true
                 } else if (tab) {
                     if (tab.empty) {
-                        maybeFocusAddressBar()
+                        focusAddressBar()
                     } else {
                         tabContainer.forceActiveFocus()
                         tab.load();
@@ -1433,18 +1433,10 @@ Common.BrowserView {
             var currentTab = tabsModel.currentTab;
             if (currentTab) {
                 if (currentTab.empty) {
-                    internal.maybeFocusAddressBar()
+                    internal.focusAddressBar()
                 } else {
                     contentsContainer.focus = true;
                 }
-            }
-        }
-
-        function maybeFocusAddressBar() {
-            if (keyboardModel.count > 0) {
-                focusAddressBar()
-            } else {
-                contentsContainer.forceActiveFocus()
             }
         }
 


UniversalSuperBox:
The jury seems split on this in other operating systems...

Browsers with new tab page content do not focus the address bar by default. For example, Chrome on Android and Safari on iOS.

Chrome on Android default new tab page -- on-screen keyboard does not show by default

Browsers with no new tab page content focus the address bar by default. For example, DuckDuckGo on Android.

DuckDuckGo on Android focuses the address bar by default

kugiigi:
Good observation. I also checked Brave on Android and it doesn't focus as well. Oh well, I guess it's really a toss up and I will give the decision to you guys 😄

I'll just leave my main argument which is that I think users open a new URL a lot more often when creating a new tab and closing the keyboard is easier than reaching the address bar, well at least on mobile....I guess those are 2 arguments 😅

UniversalSuperBox:
That appears to be the reason for the "Moving URL bar" design seen in Chrome on Android, where the URL bar is at about half-screen until clicked on. Maybe that's a better option?

cibersheep:
I myself rarely tap on the highlighted thumbnailed websites to be honest

UniversalSuperBox:
So... What should we do here? Have the keyboard appear on every new blank tab or change nothing and wait for a better design?

cibersheep:
I vote to go with the changed (focus on new tab) and review if we find a better solution or get too bad review

UniversalSuperBox (approved):
Alright, that's a decision made.

Intrinsically-Sublime:
I really wish this had not been implemented in such a forceful way, like a check box for "Focus on address bar on new tab". Myself and my wife almost never enter anything in the address bar and use the bookmarks and history almost exclusively. Now I have this keyboard blocking the clickable links and have to swipe it down which is prone to not working and miss clicking where as the old way you had to simply click in the address bar if you wanted to type and the keyboard came up.

Please add a check box to disable this (anti)feature for the rest of us. When firefox implemented focusing on the address bar instead of page content the gap was filled instantly with plugins to bring back the old way. All the other browsers I have used since then and my current Vivaldi all focus on page content not the address bar.

mateosalta:
I like it focusing, rarely use anything in the new tab page. If im opening a new tab it is to type in a new address. However in chat we were talking about why the keyboard seemed in the way for some, and noticed on other places it is dismissed if you scroll the page, I think fixing the scroll dismiss should make this better.

PR #418, opened 2020-12-08
Increased the height of tabs

### LINKED ISSUE #221: Make tabs on tablet bigger
On tablets (with the tabs on top) it seems like hitting a tab requires a bit of finesse. They are quite easily to cover with my whole finger specially as someone with fat fingers. I don't miss often but make for some odd UX. 

Can I suggest making the tabs have a bit more height, it can certainly help elder folks.

kugiigi:
Tab height in wide screen mode is too small and looks bad in desktop mode and even worse on tablets since they are hard to hit.
Tab height in narrow screen mode was also increased so that they are more visible to users and easy to spot when trying to switch to a tab.

Fixes #221

### DIFF
--- a/src/app/webbrowser/Chrome.qml
+++ b/src/app/webbrowser/Chrome.qml
@@ -82,7 +82,7 @@ ChromeBase {
                 top: parent.top
             }
             asynchronous: true
-            height: active ? units.gu(3) : 0
+            height: active ? units.gu(4) : 0
 
             Component.onCompleted: {
                 setSource(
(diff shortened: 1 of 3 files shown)


dobey (changes requested):
Please do a `git pull --rebase ubports xenial` to clean up the commits in this branch. Thanks.

kugiigi:
Fixed I think....

dobey (approved):
Looks good now. Thanks!

PR #419, opened 2020-12-08
Remove dependency on old unity-webapps-qml.

dobey:
As unity-webapps-qml seems to not really be used any more, and it also
depends on Oxide or QtWebKit which are both deprecated, just remove the
usage and dependency on it.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -77,7 +77,6 @@ Depends: ${misc:Depends},
          qtdeclarative5-ubuntu-content1,
          qtdeclarative5-ubuntu-ui-toolkit-plugin (>= 1.3) | qtdeclarative5-ubuntu-ui-toolkit-plugin-gles (>= 1.3),
          qtdeclarative5-unity-action-plugin,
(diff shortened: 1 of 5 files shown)


UniversalSuperBox (approved):

FlavioMala:
Mhhh... I don't know if it is related, but after the update to release 640 (from 637 10 dec 2020) clicking the "app" Coronavirus Italia now crashes immediately...
 BQ E5 ch dev

P.S.
Even the "radio" flas. Instead Flas (note f and F) does not crash (but maybe it remains in "loading...“); uRadio is also OK
P.P.S
Even "app" iliad (crash) too...
UBports doc too...
EU funding and Tenders... too.

balcy:
the Coronavirus Italia and UBports doc apps both do reference oxide (they do not use webapp container). I've not checked the others yet

FlavioMala:
P.P.P.S (crash):
ebay UT

PR #427, opened 2020-12-22
Fixed #426

### LINKED ISSUE #426: Swiping from the tab header selects the tab
## Steps to reproduce
1.  Swipe from the bottom to display the tabs
2. Swipe either vertically or horizontally from any of the tab headers

## Expected Results
Depending on the swipe direction, the tab should either be swipe away horizontally or scroll the view

## Actual Results
The tab is selected.

kugiigi:
Removed `press` event handler in `TabItem` and moved it in the `click` handler.

This avoids selecting a tab when starting a swipe for the tab chrome.

### DIFF
--- a/src/app/webbrowser/TabItem.qml
+++ b/src/app/webbrowser/TabItem.qml
@@ -115,15 +115,13 @@ Item {
             anchors.bottom: parent.bottom
             anchors.right: parent.right
             acceptedButtons: Qt.AllButtons
-            onPressed: {
+
+            onClicked: {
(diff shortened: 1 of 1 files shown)


UniversalSuperBox (approved):
This doesn't allow for swiping side to side, but it does allow swiping up or down. Is that what you expected? If so, I'll merge this.

kugiigi:
Yes that's the swipe I'm trying to fix but your concern is valid. Maybe I'll try to investigate on that as well since it also make sense that the user should be able to swipe from the tab to close as well.

PR #428, opened 2020-12-23
implement session history when long pressing on "back" button

lduboeuf:
screenshot20201223_163357687

### DIFF
--- a/src/app/webbrowser/NavigationBar.qml
+++ b/src/app/webbrowser/NavigationBar.qml
@@ -318,7 +318,7 @@ FocusScope {
 
             // background
             Rectangle {
-                anchors.fill: historyItemsList
+                anchors.fill: sessionHistoryContainer
                 color: theme.palette.normal.background
 
                 // border right
@@ -338,7 +338,6 @@ FocusScope {
                 anchors {
                     left: parent.left
                     right: parent.right
-                    bottomMargin: units.gu(2)
                 }
                 height: Math.min(_contentHeight, availableHeight)
                 // avoid a binding loop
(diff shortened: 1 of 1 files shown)


kugiigi:
I was supposed to work on this but I got side tracked with my other idea :smile: 
Can you also implement this in the forward button and also support right click?
I have a personal goal before the Pro1 X are shipped, make desktop use case on UT a lot more usable and Morph is main recipe for that :grin:

lduboeuf:
> I was supposed to work on this but I got side tracked with my other idea 😄

Oh ok, feel free to make it your own

> Can you also implement this in the forward button and also support right click?

ok for the forward button
Right click ?, what is it for ?

> I have a personal goal before the Pro1 X are shipped, make desktop use case on UT a lot more usable and Morph is main recipe for that 😁

Yes good example 👍

kugiigi:
Right click is for desktop use. When the user uses keyboard and mouse. It's how desktop browsers work. If I want to contribute to this PR? How do I do that? Open a PR for your branch? Working for UT is my holiday vacation 😄

lduboeuf:
> Right click is for desktop use. When the user uses keyboard and mouse. It's how desktop browsers works.

didn't know we can use the right click on desktop mode over the back and forward button, i personaly use the long press on desktop too

If I want to contribute to this PR? How do I do that? Open a PR for your branch? Working for UT is my holiday vacation 😄

Humm, i don't think you can add commits on my branch vut maybe you can create your own and i will close that one

mateosalta:
nice idea! I would go have it take up more space on the screen though, say about 3/4ths the screen on mobile portrait.

lduboeuf:
closed in favor of #444

PR #429, opened 2020-12-23
Add search tabs functionality and grid layout

### LINKED ISSUE #411: Slide to close tab is too sensitive
## Description
Swipe to close tab has been recently implemented and it's great but in my very limited experience with it, I was able to close a tab unintentionally in multiple occasions while scrolling the list of tabs

## Other comments
If there's a way to adjust sensitivity of horizontal swipes as against to vertical, then it would be good. Otherwise, maybe do not commit closing until further into the swipe.

kugiigi:
This adds a search functionality for tabs.
It can be access with `Ctrl+Space`, search button in the tab bar or swiping down from the tabs list in mobile layout.
Incidentally, this also fixed #411 

screenshot20201224_002440528
screenshot20201224_004609478

### DIFF
--- a/src/app/webbrowser/TabsList.qml
+++ b/src/app/webbrowser/TabsList.qml
@@ -44,6 +44,13 @@ Item {
 
     readonly property bool animating: selectedAnimation.running
 
+    Connections {
+        // WORKAROUND: Repeater items in listNarrowComponent stay hidden when switching from wide to narrow layout
+        // if the model is direcly assigned in its definition. This solves that issue.
+        target: browser
+        onWideChanged: if (!target.wide) searchText = " "
+    }
+
     TabChrome {
         id: invisibleTabChrome
         visible: false
@@ -140,8 +147,8 @@ Item {
                 right: parent.right
                 margins: units.gu(1)
             }
-
             placeholderText: i18n.tr("Search Tabs")
+            inputMethodHints: Qt.ImhNoPredictiveText
             primaryItem: Icon {
                 height: parent.height * 0.5
                 width: height
@@ -184,6 +191,7 @@ Item {
         sourceComponent: browser.wide ? listWideComponent : listNarrowComponent
 
         Behavior on anchors.topMargin {
+            enabled: !browser.wide
             UbuntuNumberAnimation {
                 duration: UbuntuAnimation.SnapDuration
             }
@@ -253,7 +261,7 @@ Item {
                     property real verticalMargin: horizontalMargin * ((gridDelegate.height - tabslist.tabChromeHeight) / gridDelegate.width)
 
                     title: model.title ? model.title : (model.url.toString() ? model.url : i18n.tr("New tab"))
-                    icon: model.icon
+                    tabIcon: model.icon
                     incognito: tabslist.incognito
                     tab: model.tab
                     chromeHeight: tabslist.tabChromeHeight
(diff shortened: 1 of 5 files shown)


UniversalSuperBox:
Unfortunately, this seems to conflict on TabsList.qml. I've made what I think is the proper fix for this here: https://github.com/UniversalSuperBox/morph-browser/tree/xenial_-_tabsearch... Could you check out the edited commits and edit your branch accordingly?

cibersheep:
Oooooh! I like!

UniversalSuperBox (changes requested):
Wow. What a treat to use.

But there are some concerns I have...

Sometimes when I open the tab listing, no tabs are displayed. This is easiest to reproduce when opening the browser with only one tab open. When this occurs, the following is logged:

```
file:///usr/share/morph-browser/webbrowser/TabsList.qml:256:21: Unable to assign QUrl to QQuickIcon
Ignoring sourceSize request for image url that came from grabToImage. Use the targetSize parameter of the grabToImage() function instead.
Ignoring sourceSize request for image url that came from grabToImage. Use the targetSize parameter of the grabToImage() function instead.
```

Empty mobile tab listing

It appears that the search only occurs when the Enter key is pressed. For whatever reason, I expected it to be continuous. Can the search be continuous (as the user types) without adding a lot of complexity or slowing down the model? If not, could the tab list be grayed out while typing so the user doesn't expect it to update until they press Enter? Tapping on the grayed out list should cancel the search and return to it.

kugiigi:
I haven't encountered this but I'll check
also, search should work while typing. No need to press enter

UPDATE: I immediately encountered it with 50+ tabs :sweat_smile: 
Probably an issue introduced with Qt5,12. Got to investigate further..

kugiigi:
- Disabled predictive text in the search field so that search is done as you type even with predictive text is enabled in the keyboard.

- Fixed error you mentioned about the icon which sadly isn't the cause of the blank tab list issue
- After a long battle with the issue with `Repeater`, I came up with a workaround to fix the blank tab list issue. Actually, I already have a workaround which worked fine in Qt5.9 but apparently Qt5.12 introduced another issue with that workaround wherein the tab list is blank in narrow layout until you change model data (i.e. perform search). The workaround I end up using isn't perfect but hopefully it's enough for now :sweat_smile:

UniversalSuperBox:
> The workaround I end up using isn't perfect but hopefully it's enough for now

What does this mean? What issues still remain?

kugiigi:
I can't remember anymore if there were still issues 😅
But I think what I meant there is that my workaround doesn't look good and a bit hackish. But at least it works 😄

src/app/webbrowser/TabsList.qml
```diff
@@ -18,26 +18,39 @@
...
+        searchText = ""
     }
 
     readonly property bool animating: selectedAnimation.running
 
+    Connections {
+        // WORKAROUND: Repeater items in listNarrowComponent stay hidden when switching from wide to narrow layout
+        // if the model is direcly assigned in its definition. This solves that issue.
+        target: browser
+        onWideChanged: if (!target.wide) searchText = " "
```
UniversalSuperBox, line 51: Is setting the search text to `" "` different than setting it to `""`?

UniversalSuperBox (changes requested):

src/app/webbrowser/TabsList.qml
kugiigi: yes, `""` won't fix the issue :sweat_smile: 
It's not a good workaround, I admit but that's what I came up after hours of debugging :)
Maybe I can go back in the future and see if I can find an actual fix.

PR #430, opened 2020-12-26
Fix ubports/ubuntu-touch#1344 for rotated contents on external displays

### LINKED ISSUE #1344: Webapps using the standard container are rotated in external display when in portrait aspect ratio
- Device: Nexus 5
- Channel: Devel
- Build: 20200114/2

### Steps to reproduce
1. Connect to an external display
2. Switch to desktop mode
3. Open a webapp that use the standard container. i .e Webtelegram
4. Resize the window wherein its height is bigger than its width (portrait)

### Expected behavior
The webapp will be displayed correctly.

### Actual behavior
The webapp is rotated.
screenshot20200115_044147361

kugiigi:
Set automaticOrientation to false to fix ubports/ubuntu-touch#1344

### DIFF
--- a/src/app/BrowserView.qml
+++ b/src/app/BrowserView.qml
@@ -52,6 +52,8 @@ FocusScope {
     OrientationHelper {
         id: contentsItem
 
+        automaticOrientation: false
+
         KeyboardRectangle {
             id: _osk
         }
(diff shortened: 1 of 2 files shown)


kugiigi:
I'm not exactly sure what these are for but the app works normally based on my testing on my device, clickable desktop and external display.

First commit fixes ubports/ubuntu-touch#1344
Second commit in combination with the first, fixes #445

UniversalSuperBox (approved):

PR #435, opened 2020-12-30
Bottom edge tweaks

kugiigi:
- Hide navigationbar only when the recent view is fully shown to reduce screen activities/animations when switching to the last tab (short swipe)
 - Enabled  `immediateRecognition` in the bottom edge handler to avoid scrolling webviews when swiping from the bottom
 - Removed the close button from the navigation bar
 - Fixed animation when doing a short swipe for switching to the last tab. Previous animation restarts the position from the bottom of the screen instead of continuing from the swipe/drag. It also previously does not reach the top and jumps at some point.

### DIFF
--- a/src/app/webbrowser/BottomEdgeHandle.qml
+++ b/src/app/webbrowser/BottomEdgeHandle.qml
@@ -22,7 +22,9 @@ import Ubuntu.Components 1.3
 SwipeArea {
     direction: SwipeArea.Upwards
 
-    readonly property real dragFraction: dragging ? Math.min(1.0, Math.max(0.0, distance / parent.height)) : 0.0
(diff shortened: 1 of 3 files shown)


UniversalSuperBox (changes requested):
Sounds like I'm becoming a bit of a broken record at this point... But this PR conflicts on the base branch in NavigationBar.qml. Could you fix that up? I'm not sure which part of the commit should be kept :sweat_smile:

UniversalSuperBox:
This now appears to include changes from #417 and #429. Please edit the branch so it only includes the changes needed to bring back the fullscreen hint, or recreate this PR as appropriate.

kugiigi:
All good now :+1: ...hopefully

PR #436, opened 2020-12-30
Bring back fullscreen hint

kugiigi:
Added back the fullscreen hint that shows up when the webview goes fullscreen i.e. fullscreen video

### DIFF
--- a/src/app/webbrowser/Browser.qml
+++ b/src/app/webbrowser/Browser.qml
@@ -573,7 +573,7 @@ Common.BrowserView {
         objectName: "recentView"
 
         anchors.fill: parent
-        visible: bottomEdgeHandle.dragging || tabslist.animating || (state == "shown")
+        visible: ((browser.currentWebview && !browser.currentWebview.isFullScreen) && bottomEdgeHandle.dragging) || tabslist.animating || (state == "shown")
         onVisibleChanged: {
             if (visible) {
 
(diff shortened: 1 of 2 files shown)


UniversalSuperBox:
Hmm... While the hint is correct sometimes, I'm having a lot of trouble triggering the "swipe up" action. Is it the bottom edge swipe? Can we make that clearer?

kugiigi:
> Hmm... While the hint is correct sometimes, I'm having a lot of trouble triggering the "swipe up" action. Is it the bottom edge swipe? Can we make that clearer?

Yes, it's a bottom swipe which is exactly the same as before.
Do you mean we should change the label to something like below?
`Swipe Up From The Bottom To Exit Full Screen`
or 
`Bottom Swipe To Exit Full Screen` for a shorter version.

UniversalSuperBox:
Hmm... maybe something like `Swipe from the bottom edge to exit Full Screen`... It's a weird interaction in any case, but it should do.

We could also bring this text up from the bottom of the screen rather than the top.

Other options include maintaining the bottom edge interaction method, where a hint appears for a short time followed by a chevron... but we'd need to be careful to avoid the chevron getting in the way of page content or conflicting with the real tab pip.

UniversalSuperBox:
Oh, one more thing (sorry)... On my tablet, which doesn't have a physical keyboard, the hint unhelpfully says "Press ESC to exit full screen"

kugiigi:
I moved it to the bottom and changed the text. I've also fixed issue with the text when in wide layout.
As for the chevron, I'm not sure if it's a good idea since you don't want anything shown on top of a playing video.
I feel like the hint is enough.

mateosalta:
yeah, for some reason if you try to roll your own bottom edge - not the one with cheveron/hint, it doesn't trigger as well, I think there are extra things they are doing for swipe usability too

mateosalta:
qtwebbrowser does a scroll tracker and when you drag on the screen it brings a hidden bar in sight

UniversalSuperBox:
Morph also has one -- well, webbrowser-app had one, and it got disabled when it became Morph. https://github.com/ubports/morph-browser/pull/333

UniversalSuperBox (changes requested):
This now appears to include changes from https://github.com/ubports/morph-browser/pull/417 and https://github.com/ubports/morph-browser/pull/429. Please edit the branch so it only includes the changes needed to bring back the fullscreen hint, or recreate this PR as appropriate.

kugiigi:
Yeah sorry, I combined all my open PRs into one, didn't know fixing conflicts does this :sweat_smile: 
It's good now :+1:

PR #442, opened 2021-01-04
Update browserapplication.cpp

### LINKED ISSUE #440: QtQuick Controls 2 Style isn't Suru
I saw that `QT_QUICK_CONTROLS_STYLE` is being set to `Suru` in the code however, it doesn't seem to take effect on QQC2 components. A sample I tried is `Menu` which looks like styled with the `Default` style.

Here's a sample screenshot. (Ignore the wrong orientation, it's another issue that seems to affect `Menu` :sweat_smile: )
image

balcy:
fixes https://github.com/ubports/morph-browser/issues/440
move setting of the suru style towards the top of the function

### DIFF
--- a/src/app/browserapplication.cpp
+++ b/src/app/browserapplication.cpp
@@ -185,6 +185,12 @@ bool BrowserApplication::initialize(const QString& qmlFileSubPath
         qputenv("UBUNTU_WEBVIEW_DEVTOOLS_HOST", devtoolsHost.toUtf8());
         qputenv("UBUNTU_WEBVIEW_DEVTOOLS_PORT", devtoolsPort.toUtf8());
     }
+    
+    // set suru style
+    if (qgetenv("QT_QUICK_CONTROLS_STYLE") == QString())
+    {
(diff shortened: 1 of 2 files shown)


balcy:
hint for test:

- open "Settings" from the main menu

the selection for the default zoom should look like this (the same shape with dark colors in dark mode):
image

it should NOT look like this:
image

kugiigi:
Looks like `TabPreview` has to be updated for this.
It uses `SwipeDelegate` which has padding in `Suru`.
The fix is to set the following. Apparently, `padding: 0` isn't enough.
```
leftPadding: 0
rightPadding: 0
topPadding: 0
bottomPadding: 0
```

balcy:
ok thanks for the hint!

dano6:
Maybe we should define QT_QUICK_CONTROLS_STYLE=Suru systemwise in utouch, because currently Suru is not the default QQC2 theme.

UniversalSuperBox (approved):

PR #444, opened 2021-01-05
Adds navigation history in the back/forward button

kugiigi:
Press-and-hold and right-click on the back and forward buttons will open a popup with the navigation history just like on other browsers. Currently, it uses `Popover` with custom background color since it doesn't "popup" enough on white contents which websites usually are. I will try to use Popup/Menu from QQC2 and see if it looks better.

This is an expansion of the initial effort in #428 . @lduboeuf  and I already talked about this :smiley:


kugiigi:
This needs #430 otherwise, the popup is rotated.
#442 would also make the style look better.

kugiigi:
Here's some screenshots
screenshot20210108_012533836
screenshot20210108_012600989
screenshot20210108_012658893

UniversalSuperBox (changes requested):
Erm... Well... I guess I have to say it again.

This PR has a merge conflict.

UniversalSuperBox (approved):

PR #464, opened 2021-02-18
enable html5 notifications

balcy:


src/app/NotificationsAccessDialog.qml
```diff
@@ -0,0 +1,61 @@
...
+
+Dialog {
+    id: dialog
+    
+    property string securityOrigin
+    property bool showRememberDecisionCheckBox
+    modal: true
+
+    title: i18n.tr("Permission Request")
+    text: securityOrigin + "<br>" + i18n.tr("This page wants to know create notifications.")
```
UniversalSuperBox: I don't think "know create notifications" is proper for this case. This will also require a potfile update.

src/app/notifications-proxy.cpp
```diff
@@ -0,0 +1,77 @@
...
+    PushClient::instance()->updateCount();
+}
+
+QJsonObject NotificationsProxy::buildMessage(const QString & tag, const QUrl & origin, const QString & title, const QString & body) const
+{
+    QJsonObject notification;
+    notification["tag"] = tag;
+    notification["card"] = buildCard(origin, title, body);
+    notification["sound"] = false;
+    //notification["vibrate"] = vibrate();
```
UniversalSuperBox: Any specific reason for adding commented code to the codebase?

UniversalSuperBox (changes requested):
Hmm. This seems like a huge review without much ability for me to give useful feedback. But I did find a couple of things.

src/app/NotificationsAccessDialog.qml
balcy: should have been "wants to create notifications", a copy and paste error from "wants to know your location" I think.

src/app/notifications-proxy.cpp
balcy: no not really, I guess it was like that in the example I found

UniversalSuperBox (changes requested):
Thanks, that's the static analysis done.

Note that this **DOES NOT** enable webpush. At all. Attempting to use [the Microsoft Edge Web Push demo](https://webpushdemo.azurewebsites.net/) throws the error `Registration failed - push service not available` in the browser console.

Instead, this only enables the [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API). The notifications API allows an active tab to send a notification via JavaScript. If a tab becomes suspended due to memory pressure, it will be unable to send notifications any more. This limits the feature's usefulness for now, but could lead to better things in the future.

I tested the feature with https://www.davidwalsh.name/demo/notifications-api.php.

When a website tries to show a notification, a permission request appears.

Permission Request: https://davidwalsh.name/; this page wants to create notifications. Checkbox to remember decision. Provided options: Allow, Deny

If allowed, the single notification that the website tried to send appears.

Notification title: "Title". Notification body: "I am the body text!"

If the website tries to send another notification, the permission request appears again. I believe this should be changed. It is not useful to be asked whether to allow or deny a push notification every time a website tries to send it... you might as well have just allowed the notification in the first place. Either the "Remember decision" checkbox should be checked by default or, probably more useful, removed and the options changed to "Always Allow" and "Always Deny"

UniversalSuperBox:
Also interesting: Tapping on a notification opens the website attached to the notification in a new tab. I think that the browser should switch to the site's open tab instead of opening a new tab. That's the behavior of other browsers, at least.

balcy:
for the first issue (authentication asked twice): 
Was it during the same session (without closing morph), that the page asked for the confirmation again ? I thought it would remember it like the audio permission.

I think getting rid of the "allow permanently" checkbox is not possible, because in incognito mode it is not enabled because no permanent data about domains is saved then.

The second one seems a bit trickier: We would need to change the external URL handling and not open it in another tab, if there is already an open tab with that URL. You're right other browser do it like that.

UniversalSuperBox:
> Was it during the same session (without closing morph), that the page asked for the confirmation again?

Yes. I didn't leave the page. I just clicked the button to send a notification, said "no" to the prompt, then clicked the button to send a notification again.

> I think getting rid of the "allow permanently" checkbox is not possible, because in incognito mode it is not enabled because no permanent data about domains is saved then.

That seems fine. Firefox silently changes this behavior in Private Mode, the "Allow" button changes from enabling "Allowed" to "Allowed temporarily".

> We would need to change the external URL handling and not open it in another tab

True. I think it's required, though. Take Firefox on https://www.davidwalsh.name/demo/notifications-api.php as an example: if you close the tab, the notification is removed from Gnome! The notifications aren't meant to outlive a tab, and they aren't meant to have any other context than that.

balcy:
ah ok, that's true the "no" was not saved, but the "yes" is.
Ok that should now work with commit https://github.com/ubports/morph-browser/pull/464/commits/ca01e97c53e2db798e073a1f3a67d5658fa7536c

For the last issue (focus tab instead of opening the link in the new tab):
On webapps it is already correct I think, because the app is opened (or focused if already open)

For morph as the default URL handler we could generally change the logic like that:
- Does a tab with that URL exist ? -> focus that tab, otherwise open the URL in a new tab.

Would it be a problem, if all URL links (e.g. links clicked in apps) are handled that way by morph-browser ?

balcy:
that would be different to an URL passed on the desktop via `firefox <url>` (if firefox is already running) or `sensible-browser <url>` : Those open a new tab with that URL

balcy:
ok with the latest commit, morph switches to the (already open) tab with the url from the notification

PR #467, opened 2021-02-18
Fix some tab previews are not visible in 5.12

### LINKED ISSUE #457: Preview of tabs sometimes are gone when switching between tabs
### Steps to Reproduce
1. Switch to the second tab
2. Switch to the 3rd tab
3. Open the tab switcher

### Expected Results
Previews of all tabs you switched to are displayed

### Actual Results
Preview of the 3rd tab in the list is blank and shows an error in the log
```...TabPreview.qml:122:13: QML QQuickImage: Protocol "itemgrabber" is unknown```

### Additional Information
- Reopening the app restores the missing preview
- It seems that this error was introduced in Qt5.12
- It seems to only happen with previews created with `grabToImage` and whenever the preview is unloaded in the view due to scrolling or model change.

dano6:
This should fix https://github.com/ubports/morph-browser/issues/457


dobey (approved):
Looks OK to me.

UniversalSuperBox (approved):

PR #471, opened 2021-03-14
Remove unused libapparmor dependency

nicolasfella:
The CMake variables are empty


dobey (changes requested):
I don't think this is correct. Instead we should actually use apparmor to check the confinement profile, as was done previously, though however was commented out. I've made a PR to bring the check back at https://github.com/ubports/morph-browser/pull/473

nicolasfella:
Fair enough

PR #472, opened 2021-03-15
rollback close button

mateosalta:
close button in header bar is unused now, now all the close buttons are shown on tabs, the tab close on the navigation bar flashes while the tab scrolls out of view.


kugiigi:
This is also covered by #435  if ever that PR gets merged :)
I wasn't able to remove the signal `closeTabRequested` though :sweat_smile:

mateosalta:
ah, awesome!

PR #486, opened 2021-05-03
Update MorphSharedWebContext.qml

balcy:
define a storageName for the non-incognito profile.
with QtWebEngine 5.15.2 the browser no longer remembered cookies, e.g. logins (without storageName defined)

### DIFF
--- a/src/Morph/Web/MorphSharedWebContext.qml
+++ b/src/Morph/Web/MorphSharedWebContext.qml
@@ -27,7 +27,7 @@ QtObject {
     property QtObject sharedContext: Morph.WebContext {
         id: context
         offTheRecord: false
-        storageName: "MorphSharedWebContext"
+        storageName: "Default"
     }
 
     property QtObject sharedIncognitoContext: Morph.WebContext {
--- a/src/Morph/Web/MorphSharedWebContext.qml
+++ b/src/Morph/Web/MorphSharedWebContext.qml
@@ -27,6 +27,7 @@ QtObject {
     property QtObject sharedContext: Morph.WebContext {
         id: context
         offTheRecord: false
+        storageName: "MorphSharedWebContext"
     }
 
     property QtObject sharedIncognitoContext: Morph.WebContext {
--- a/src/app/webcontainer/webapp-container.qml
+++ b/src/app/webcontainer/webapp-container.qml
@@ -170,6 +170,7 @@ BrowserWindow {
         onLoaded: {
             var context = item.currentWebview.context;
             context.offTheRecord = false;
+            context.storageName = "WebappContainerContext";
             onlineAccountsController.setupWebcontextForAccount(context);
             item.currentWebview.settings.localContentCanAccessRemoteUrls = localContentCanAccessRemoteUrls;
 
--- a/src/app/webcontainer/webapp-container.qml
+++ b/src/app/webcontainer/webapp-container.qml
@@ -170,7 +170,7 @@ BrowserWindow {
         onLoaded: {
             var context = item.currentWebview.context;
             context.offTheRecord = false;
-            context.storageName = "WebappContainerContext";
+            context.storageName = "Default";
             onlineAccountsController.setupWebcontextForAccount(context);
             item.currentWebview.settings.localContentCanAccessRemoteUrls = localContentCanAccessRemoteUrls;
 


UniversalSuperBox:
While this should be merged, are you considering carrying forward https://gitlab.com/ubports/core/packaging/qtwebengine/-/blob/ubports/xenial/debian/patches/0014-maintain-offTheRecord-default.patch?

This will maintain the old behavior for apps which use QtWebEngine directly instead of through Morph.Web.

UniversalSuperBox:
Additionally, will this make users sign in to webapps again? I think this changes which subdirectory on disk is used for saving cookies and site data.

balcy:
The prototype branch QtWebEngine 5.15.2 still does contain the patch: https://gitlab.com/balcy/qtwebengine/-/blob/xenial_-_prototype5-15-2/debian/patches/0009-maintain-offTheRecord-default.patch

For some reason, leaving the "storageName" empty on QtWebEngine 5.15.2 now results in persistentCookiesPolicy = WebEngineProfile.NoPersistentCookies, even if both persistentStoragePath is set and persistentCookiesPolicy is set to another value. With the storageName set persistent cookies are possible, I've not really understood why exactly.

UniversalSuperBox:
Okay. Will this cause users to need to sign in to all their webapps again, though?

balcy:
I've tried with a web app QtWebEngine 5.14.2:
- not logged in (e.g. github), storageName not set
- logged in
- still logged in on next app start
- added context.storageName = "WebContainerContext" to webapp-container.qml
- still logged in on next app start

maybe you can try it as well by changing webapp-container.qml directly on the device.

But when switching from QtWebEngine 5.14.2 -> 5.15.2  or 5.15.2 -> 5.14.2 I had to login again each time.

UniversalSuperBox:
Do you mean that *with this PR installed*, you had to log in to your online services when switching between QtWebEngine versions?

balcy:
actually both ways, but without the PR for me the profile of QtWebEngine 5.15.2 has the "no persistent cookies" policy

- without the PR installed, when switching QtWebEngine 5.14.2 -> 5.15.2 all logins were lost.
- they were not remembered at all from this point with 5.15.2, e.g. you had to login each time morph started, as if it was incognito mode, even if it was the "offTheRecord=false" profile.

- since I found no other solution for this, I've added the storageName (still not sure why the functionality depends on it)
- with that set, the login is kept in 5.15.2

Then with the PR installed:
- when going back to 5.14.2 the logins are lost once
- when going back to 5.15.2 the logins are lost once

UniversalSuperBox:
Losing a user's logins on an OTA update seems like a huge problem... Do you think we'll be able to fix that?

balcy:
ok I think by using "Default" as the storageName (recent commits), the sessions are not lost on the QtWebEngine upgrade.
E.g. the PR could be already merged now and then morph and webapp-container are prepared for QtWebEngine 5.15.2.

Apps using QtWebEngine directly and not setting the storageName would likely behave like "offTheRecord=true" with QtWebEngine 5.15.2.

https://doc.qt.io/qt-5/qwebengineprofile.html#details (the C++ version) contains the following description:
__Using_ QWebEngineProfile() without defining a storage name constructs a new off-the-record profile that leaves no record on the local machine, and has no persistent data or cache._

So maybe it is really the intention of the QtWebEngine developers to implement "privacy by default" here.

I understand that the "offTheRecord" patch of QtWebEngine was introduced in order to not break the app behavior for users and developers within our eco system.

At that time (Sept. 2020) we switched from QtWebEngine 5.11 to 5.14, but the change of the default "offTheRecord=true" existed since QtWebEngine 5.12.0 (December 2018).

There are 2 examples (scenarios) that make me think it is better to stick with upstream here.

1) An ubuntu touch app (based on QtWebEngine) saves the session on the device an works fine without any profile settings (offTheRecord / storageName). At some point the developer wants to use the same codebase (or as much as possible) on another platform. He / she will wonder, why suddenly it only saves the data in memory, and no longer on the disk. Even if on this system there is exactly the same QtWebEngine version, but the patched one on Ubuntu Touch.

2) A developer has an app that has no intention and need to save browser-related data on the disk. It depends on QtWebEngine and is not available for Ubuntu Touch yet. The developer has verified that no QtWebEngine data (browser session / cache) is saved on the disk, and does not set "offTheRecord" and "storageName", because he / she is used to this being the default for as long as they can think. When they port the app to ubuntu touch, there it will suddenly start creating a permanent profile, which the dev does not expect, and which might be a privacy issue for him or the users, and possibly not noticed immediately. Of course he/she should test this, but why would they think QtWebEngine on Ubuntu Touch behaves differently ?

IMO we only should change what we are in charge of, e.g. Morph.Web, webapp-container, morph-browser, ...  but not an upstream package like QtWebEngine. (For default behavior, (backporting of) bugfixes is fine of course)

UniversalSuperBox:
Yeah, if we can give app developers ample time and notification of changes, we can make offTheRecord's default change. The problem in the previous case was how close we were to release and how little heads-up we could give to app devs.

UniversalSuperBox (approved):
If the Default argument works, I'm happy with the listed changes. Let's do it.

balcy:
ok fingers crossed :  )

Issue #510, opened 2021-08-08
URLs that contain an encoded slash character (%2F) are not handled correctly


balcy:
- navigate to URL https://ci.ubports.com/blue/pipelines
- search morph-browser
- click on the result
- the current page should now be https://ci.ubports.com/blue/organizations/jenkins/ubports%2Fmorph-browser/activity

- click on the address bar
- press enter

Expected Result: 
The page should reload, the URL should not change
Actual result:
- The URL changes to https://ci.ubports.com/blue/organizations/jenkins/ubports/morph-browser/activity
- you see "Page not found (404)

The same issue is when the browser is restored:
- select "restore previous session at startup" in settings
- keep a tab with https://ci.ubports.com/blue/organizations/jenkins/ubports%2Fmorph-browser/activity open
- on the next start the URL is https://ci.ubports.com/blue/organizations/jenkins/ubports/morph-browser/activity

Another instance:
- copy the URL https://ci.ubports.com/blue/organizations/jenkins/ubports%2Fmorph-browser/activity to the clipboard
- set a custom homepage (in settings) with that pasted URL
- the saved custom homepage is actually https://ci.ubports.com/blue/organizations/jenkins/ubports/morph-browser/activity

(%2F was replaced by /)

balcy:
in those cases the QML url type is used, which works well when initialized from a QUrl, and in most cases when initialized from a string, but there is an exception with the percent encoded slash (%2F)

https://doc.qt.io/qt-5/qml-url.html

> Additionally, URLs may contain encoded characters using the 'percent-encoding' scheme specified by RFC 3986. These characters will be preserved within properties of type url, to allow QML code to construct precise URL values. An exception to this rule is the preemptive decoding of directory-separator characters ('/') - these characters are decoded to allow the URL to be correctly classified.

The proposed solution is to use the newly created UrlUtils class (C++ backend) instead of UrlUtils.js

PR #511, opened 2021-08-08
replace UrlUtils.js by UrlUtils class in the backend

balcy:


balcy:
fixes https://github.com/ubports/morph-browser/issues/510

PR #517, opened 2021-09-09
Add reader mode

DavidVentura:
Hi!
This PR adds "Reader mode" to morph -- the readability.js file is taken straight out of Mozilla's repo [here](https://github.com/mozilla/readability). The style and header are the bare minimum to get something usable, they could be improved.

I have absolutely no idea if the placement of my code makes any sense -- to me the 'readerMode' state should be stored in each tab, but that's likely incorrect

Attached, a video of behavior

https://user-images.githubusercontent.com/3650670/132724146-9555f134-5a55-45f7-bbf3-97a9ae66dfe9.mp4


src/app/webbrowser/TabComponent.qml
```diff
@@ -279,5 +296,15 @@ Component {
                 }
             }
         }
+        function toggleReaderMode() {
+            if (browserTab.readerMode) {
+                browserTab.reload()
```
DavidVentura, line 301: i thought about storing the entire dom in the `makeDomReadable` function, but I'm not sure it's worth the tradeoff.
I don't really understand how that'd work with something like a video that's playing (if a reference is stored in a "shadow dom", does it keep playing? etc)

mateosalta:
i think eventually we will switch over to the builtin blink overlay scrollbars, the css injection scrollbar theme is a bit hacky and not as nice, so probally dont worry about them too much unless you know an easy way to make an exception for that script.

mateosalta:
I think the placment makes sense to me - but perhaps there will be people who want less items in the navigation? with that in mind it may make sense to have a disable option in settings

mateosalta:
try looking at 'stock-ebook' , 'note'  icons from our suru theme

DavidVentura:
stock_ebook is indeed nicer, thanks!

PR #519, opened 2021-10-11
Update ar.po

herobrine7gamer:


herobrine7gamer:
Why is the translation not accepted? is the translation wrong? or is it the qt format? please let me know to solve the problem.

ivoxavier:
@herobrine7gamer translations are managed by weblate: https://translate.ubports.com/projects/ubports/

Thank you

